PR #38, opened 2022-09-27
Flip logic for choosing ingress

esphen:
This ensures other external patterns are implicitly supported, like `norgeskart.no`. The only DNS wildcard that is currently pointing to the internal ingress is `*.env.skip.statkart.no`.

The only tradeoff is that the implicit behavior makes it possible to expose things by accident, but I see this as unlikely as that would require pointing DNS to the external dns first.

### DIFF
--- a/controllers/ingress_gateway.go
+++ b/controllers/ingress_gateway.go
@@ -66,10 +66,10 @@ func (r *IngressGatewayReconciler) Reconcile(ctx context.Context, req reconcile.
 				return err
 			}
 
-			if isExternal(hostname) {
-				gateway.Spec.Selector = map[string]string{"ingress": "external"}
-			} else {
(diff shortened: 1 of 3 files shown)


Flexman94 (approved):

BardOve (approved):

aa-kv (approved):
Changed reference from `[^.]\.kartverket\.no` to `[^.]\.skip\.statkart\.no` effectively disables subdomains on the `kartverket.no` domain.

esphen:
> Changed reference from `[^.]\.kartverket\.no` to `[^.]\.skip\.statkart\.no` effectively disables subdomains on the `kartverket.no` domain.

It shouldn't, the behavior is the same as before with subdomains of kartverket.no being assigned to the external gateway.

PR #89, opened 2022-11-17
[SKIP-703] Manage resources in Skiperator containers

anderssonw:
Sets Requests required for CPU and Memory, and Limit required for Memory

Adds some extra documentation regarding testing and restarting after making changes

Bumps run-terraform to newest version to remove set-output warning

Adds new test for asserting that the container will not have `resources.limits.cpu` set even though the application CRD does, and updates all the old tests with new required fields


api/v1alpha1/application.go
```diff
@@ -189,3 +222,40 @@ func max[T constraints.Ordered](a, b T) T {
...
+	if !resList.StorageRequest.IsZero() {
+		resourceList["storage"] = resList.StorageRequest
+	}
+
+	if !resList.EphemeralStorageRequest.IsZero() {
+		resourceList["ephemeral-storage"] = resList.EphemeralStorageRequest
+	}
+
+	return resourceList
+}
```
anderssonw, line 261: This feels like a dumb solution, but I wasn't sure if it would be possible to use the built in kubebuilder validation if I didn't make it two different structs.

controllers/deployment.go
```diff
@@ -135,7 +135,20 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context, application *skipe
 			}
 		}
 
-		container.Resources = application.Spec.Resources
+		container.Resources.Limits = application.LimitToCoreResourceList(application.Spec.Resources.Limits)
+		container.Resources.Requests = application.RequestToCoreResourceList(application.Spec.Resources.Requests)
```
esphen, line 139: We probably need a nil check here. `application.Spec.Resources` is set to required, but there might be some resources on the cluster that have not defined this today which will make this run into a nil pointer exception.

api/v1alpha1/application.go
```diff
@@ -3,35 +3,68 @@ package v1alpha1
...
+// +kubebuilder:object:generate=true
+type LimitResourceList struct {
+	//+kubebuilder:validation:Optional
+	CpuLimit resource.Quantity `json:"cpu,omitempty"`
+	//+kubebuilder:validation:Required
+	MemoryLimit resource.Quantity `json:"memory"`
+	//+kubebuilder:validation:Optional
+	StorageLimit resource.Quantity `json:"storage,omitempty"`
+	//+kubebuilder:validation:Optional
+	EphemeralStorageLimit resource.Quantity `json:"ephemeralStorage,omitempty"`
```
esphen, line 44: Should we support `EphemeralStorageLimit` and `StorageRequest`?

api/v1alpha1/application.go
```diff
@@ -3,35 +3,68 @@ package v1alpha1
...
 type ApplicationSpec struct {
 	//+kubebuilder:validation:Required
 	Image string `json:"image"`
 	//+kubebuilder:validation:Optional
 	Command []string `json:"command,omitempty"`
 
-	//+kubebuilder:validation:Optional
-	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+	//+kubebuilder:validation:Required
+	Resources *ResourceRequirements `json:"resources"`
```
esphen, line 67: Note that this change needs to be advertised in #gen-skip when it gets merged. If anyone has defined resources today it might mean that they are unable to deploy any new code before they add that field because the validation says it's required.

tests/cpu-limit-unset/00-errors.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Deployment
+metadata:
+  name: test
+spec:
+  template:
+    spec:
+      containers:
+        - resources:
+            limits:
+              cpu: 10m
```
esphen, line 11: How does this file assert that the test should fail?

esphen (commented):
Nice work :slightly_smiling_face:
A few comments and questions below

tests/cpu-limit-unset/00-errors.yaml
anderssonw: From the [kuttl documentation](https://kuttl.dev/docs/testing/asserts-errors.html#format) they state that anything present in the `index-errors.yaml` file will cause a test error. As the CR removes the CPU limit even if it is present in the CRD spec, it should never be present.

We tested with the other resources, and those cause the test to fail, so this was the simplest way to check if a field was *not* present.

api/v1alpha1/application.go
anderssonw: Is it currently unsupported?

esphen: Not documented, so not explicitly supported at least

anderssonw: if you prefer i could remove these. thought if they're not currently unsupported i dont see why we should, in case they might be useful for some case

esphen (approved):

berg-thom (approved):

PR #93, opened 2022-11-23
nil check on configmap to prevent crash

berg-thom:
Added a nil check on the instanaConfigMap to prevent skiperator from crashing when the CIDR block is not present, for example when testing locally.


KronenE (approved):
Feilen på test er pga nedetid på testclusteret så ignorerer den

PR #99, opened 2022-12-02
Forcibly override field manager for clusterrole too

anderssonw:
Fixes faulty deploy in develop env due to local skiperator run in the wrong context. See #98 for solution to the underlying problem.


Flexman94 (approved):
I

KronenE (approved):

PR #102, opened 2022-12-02
[SKIP-768] Resource management fixes

anderssonw:
Awaiting a response:

https://github.com/hashicorp/terraform-provider-kubernetes/issues/1923


anderssonw:
Outdated. According to the linked issue the issue was how we updated fields of an API.

Not a huge benefit for potentially a lot of work here

PR #103, opened 2022-12-05
Fix order of arguments

omaen:
I put the -d flag by mistake after -t which expects a string argument.

Set the -t flag and value as a single argument instead.


anderssonw (approved):
fort gjort 😩

KronenE (approved):

PR #106, opened 2022-12-08
[SKIP-771] Safer deployment

anderssonw:
Makes the regular deployment workflow sequential instead of parallell. Also creates a new workflow for sequential deployment of a specific image, allowing for quicker rollbacks to old versions, should that be necessary.

### DIFF
--- a/.github/workflows/dispatch-deploy.yml
+++ b/.github/workflows/dispatch-deploy.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       image_tag:
-        description: "Tag/Digest of image to deploy. This image should be a well-known working version of Skiperator. Potential image digests are found under skiperator-controller packages."
+        description: "Tag of image to deploy. This image should be a well-known working version of Skiperator. Potential image tags are found under skiperator-controller packages."
         required: true
         type: string
 
--- a/.github/workflows/dispatch-deploy.yml
+++ b/.github/workflows/dispatch-deploy.yml
@@ -1,4 +1,4 @@
-name: Deploy Skiperator Unsafe, Quick
+name: Deploy Skiperator Parallell With Dispatch, Quick/Unsafe
 
 on:
   workflow_dispatch:
--- a/.github/workflows/dispatch-deploy.yml
+++ b/.github/workflows/dispatch-deploy.yml
@@ -3,8 +3,8 @@ name: Deploy Skiperator Parallell With Dispatch, Quick/Unsafe
 on:
   workflow_dispatch:
     inputs:
-      image:
-        description: "Image to deploy. This image should be a well-known working version of Skiperator. Image digest found under skiperator-controller"
+      image_tag:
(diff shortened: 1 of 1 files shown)


.github/workflows/dispatch-deploy.yml
```diff
@@ -0,0 +1,86 @@
...
+      id-token: write
+      pull-requests: write
+      contents: read
+    uses: kartverket/github-workflows/.github/workflows/run-terraform.yml@v3.1.0
+    with:
+      runner: ubuntu-latest
+      environment: sandbox
+      kubernetes_cluster: atkv1-sandbox
+      terraform_workspace: sandbox
+      terraform_option_1: -var=image=ghcr.io/kartverket/skiperator-controller:${{ inputs.image }}
```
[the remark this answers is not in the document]
anderssonw: 💯

Flexman94 (approved):

.github/workflows/dispatch-deploy.yml
```diff
@@ -0,0 +1,86 @@
...
+      id-token: write
+      pull-requests: write
+      contents: read
+    uses: kartverket/github-workflows/.github/workflows/run-terraform.yml@v3.1.0
+    with:
+      runner: ubuntu-latest
+      environment: sandbox
+      kubernetes_cluster: atkv1-sandbox
+      terraform_workspace: sandbox
+      terraform_option_1: -var=image=ghcr.io/kartverket/skiperator-controller:${{ inputs.image_tag }}
```
odasko, line 24: After a `:` it is only possible to specify a tag. If you want to specify a digest you will need to use a `@` instead. 
(i.e. the image formats are `registry/repository:tag` and `registry/repository@digest`).
The description on line 7 is a little misleading in that regard

anderssonw: But we do not use tags in the main deploy function, do we? We use `${{ needs.build.outputs.version }}` and I'm not sure I understand where we specify a version in the package: https://github.com/kartverket/skiperator/pkgs/container/skiperator-controller/57846931?tag=sha-92381fd8f7773ac447cb9cf07693a92a1d777da1

@oda

anderssonw: @odasko opsi

odasko: For the most recent package her: https://github.com/kartverket/skiperator/pkgs/container/skiperator-controller 
You can either specify the image using the tag like this `ghcr.io/kartverket/skiperator-controller:sha-92381fd8f7773ac447cb9cf07693a92a1d777da1` OR you can specify the image using the digest like this `ghcr.io/kartverket/skiperator-controller@sha256:b60b083a820afc92ca425c2a5424c8d25a76f83ca084cb5722001009edfa37d1`
Notice the use of `:` when using the tag and `@` when using the digest

odasko: If you look her the version output is the same as the tag 
Screenshot 2022-12-08 at 15 16 51

anderssonw: aha, cheers :)

PR #116, opened 2022-12-20
[SKIP-764] Fix string filters

anderssonw:
This PR aims to improve some filters put on a few different network services. Currently they were split into segments by using `-` as a delimiter, which did not work whenever teams had multiple dashes in their application name.

The filter has been improved to use regex instead of segment splitting on a delimiter. There still exists one segment split filter, which I have been unable to improve so far due what it is attempting to fetch from the segments:
image

It is used in the main reconcile function to watch over specific certificates:
image

Presumably, this is not the best way of handling this, and we could probably use some sort of label selector or similar instead.

As of now, this change makes Skiperator actually reconcile any gateway/serviceEntry/virtualService in which the application name contains dashes, which was not the case earlier.

Due to the last string splitting, it will not reconcile certificates where this is the case. Due to this probably requiring a bigger rewrite of some reconciliation code, this will be contained in a different PR.


aa-kv (approved):

controllers/application/ingress_gateway.go
```diff
@@ -109,13 +109,9 @@ func (r *ApplicationReconciler) reconcileIngressGateway(ctx context.Context, app
 	return reconcile.Result{}, err
 }
 
-// Filter for gateways named like *-ingress
+// Filter for gateways named like *-ingress-*
 func isIngressGateway(gateway *networkingv1beta1.Gateway) bool {
-	segments := strings.Split(gateway.Name, "-")
+	match, _ := regexp.MatchString("^.*-ingress-.*$", gateway.Name)
```
esphen, line 114: Is it possible to write a test for verifying that it no longer creates and re-creates gateways like a madman?

esphen (commented):
100% agree that this should be solved with labels in the future instead of parsing names.

In the next iteration we could determine a standard for labels assigned to SKIP-created resources. We already have the `skip.kartverket.no/description` metadata, introducing more like `skip.kartverket.no/component` means we should document their usage. We could also consider the [standard `app.kubernetes.io/component` label](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) for this.

controllers/application/ingress_gateway.go
anderssonw: Im looking into some more advanced tests in #118.

This PR really only allows applications to have dashes in their names, and does not change the creation/re-creation of madman-gateways. Should be fixed in #118 though :)

esphen (approved):

PR #118, opened 2022-12-21
[SKIP-828] Improve networking resource reconciliation loops

anderssonw:
Managed active field was created per application, and therefore every other namespaces ingresses, gateways and certificates were removed due to not finding these as active. This logic has been rewritten to not keep a list of active resources, but rather determine which resources is active by other means, such as ownership and application spec.

Certificates are now reconciled, but some issues are still present, as there is no ownership on the cert. That means a deletion event on an application will not delete certs. #117 will handle this


KronenE (approved):

anderssonw:
> Is it possible to write tests to verify that the correct certificates are deleted? Since this is mainly based on name-matching it would be nice to catch if the convention is changed later.

Good point. I will take a gander 💪

anderssonw:
Working on tests here, making this a draft for the time being.

controllers/application/certificate.go
```diff
@@ -62,33 +56,34 @@ func (r *ApplicationReconciler) reconcileCertificate(ctx context.Context, applic
...
+		if certificateInApplicationSpec {
 			continue
 		}
 
-		// Skip active certificates
-		_, ok := active[certificate.Name]
-		if ok {
+		// We want to delete certificate which are not in the spec, but still "owned" by the application.
+		applicationNameInCertificate := strings.Contains(certificate.Name, application.Name)
+		applicationNamespaceInCertificate := strings.Contains(certificate.Name, application.Namespace)
```
esphen, line 80: Her ser det ut som du kan gå på en smell om f.eks. namespacet heter dsa og applikasjonen heter dsa-frontend. Siden den kun sjekker om strengen finnes i navnet uavhengig av hva det er vil du potensielt få uventet oppførsel hvor en applikasjon vil matche på andre applikasjoners certificates siden de inneholder navnet på namespacet i application.name slik at `applicationNameInCertificate` blir true. 

Bedre å bruke regex her?

anderssonw: jeg tror ikke det ville forårsaket en feil i namespace: dsa og name: dsa-frontend tilfellet, men ja det vil nok være lurt å matche på hele namespace-name over individuelt, ettersom begge kan inneholde hverandre i forskjellige tilfeller

esphen (approved):

PR #151, opened 2023-02-07
[SKIP-953] Exclude namespaces using a configmap

anderssonw:
This PR attempts to use a ConfigMap called `namespace-exclusions` in the `skiperator-system` namespace to avoid reconciling certain namespaces. This is a lot more agile than having to add and redeploy skiperator for every change we want to make to this list.

The commit a5a43d5 was added simply because I started working in that folder before realising that would not work. Can be removed from PR if wanted, but its somewhat of a clean-up so I let it stay.

I am not sure if we want to keep the static list in the code or if we want to exclusively use the config map. In addition, I'm not sure if we should keep an example of this config map in the repo, and where it should stay if that is the case. Awaiting feedback here :)

It would be preferable to use the `WithEventFilter` function on the manager, as earlier, but you are not allowed to fetch kubernetes resources before the controllers are setup. As such, I'm not sure this approach is possible.


controllers/namespace/exampleExclusionCMap.yaml
```diff
@@ -0,0 +1,32 @@
...
+  vault: exclude
+  sysdig-agent: exclude
+  sysdig-admission-controller: exclude
+  instana-agent: exclude
+  kubecost: exclude
+  argocd: exclude
+  crossplane-system: exclude
+  upbound-system: exclude
+  instana-autotrace-webhook: exclude
+  fluentd: exclude
```
anderssonw, line 32: her hadde man strengt tatt kun trengt å inkludere en nøkkel og ikke bry seg om verdien, men om man gjør det på følgende måte så blir det kanskje vanskeligere å gjøre feil. vet ikke om det er hensiktsmessig.

controllers/namespace/controller.go
```diff
@@ -19,15 +23,76 @@ type NamespaceReconciler struct {
...
+		err = r.GetClient().Create(ctx, &namespaceExclusionCMap)
+		if err != nil {
+			fmt.Printf("Something went wrong creating namespace-exclusion config map: %v", err.Error())
+			return true
+		}
+	}
+	if err != nil {
+		fmt.Printf("Something went wrong getting namespace-exclusion config map: %v", err.Error())
+		return true
+	}
```
anderssonw, line 80: should we return true or false upon errors here? should we maybe use `panic()` instead?

anderssonw:
Also worth mentioning: Adding a namespace to either the in-code list or the ConfigMap will _not_ delete resources created by namespace reconciliation. This must be done manually. I'm not entirely sure how we should go about deleting, perhaps someone else might know how.

controllers/application/configmap.go
```diff
@@ -25,32 +25,46 @@ type CredentialSource struct {
...
+		if errors.IsNotFound(err) {
+			r.GetRecorder().Eventf(
+				application,
+				corev1.EventTypeWarning, "Missing",
+				"Cannot find configmap named gcp-identity-config in namespace skiperator-system",
+			)
+		} else if err != nil {
+			r.SetControllerError(ctx, application, controllerName, err)
+			return reconcile.Result{}, err
+		}
```
esphen, line 47: Consider creating a `getConfigMap` utility function?

controllers/namespace/exampleExclusionCMap.yaml
esphen: Seems it has to be a string-to-string map. The only alternative would be to put an array in a string and decode it as JSON in code, but that has its own complexity, so I guess this is fine

pkg/util/helperfunctions.go
```diff
@@ -14,44 +11,6 @@ func IsInternal(hostname string) bool {
...
-	// TODO PoC, add NetworkPolicies after
-	"sysdig-agent",
-	"sysdig-admission-controller",
-	"instana-agent",
-	"kubecost",
-	"argocd",
-	"crossplane-system",
-	"upbound-system",
-	"instana-autotrace-webhook",
-	"fluentd", //POC
```
esphen, line 48: Are we going to create these on the clusters with terraform by creating files in the deploy directory?

controllers/namespace/controller.go
esphen: I think this is a panic, as it is impossible for skiperator to continue if it does not know what namespaces are excluded

pkg/util/helperfunctions.go
anderssonw: I'm sort of torn on how we should do this in general.

On the one hand, we want to be able to exclude namespaces somewhat more agile than what we do now. On the other hand, if the discrepancy between all the ConfigMaps between environments differ a lot, thats also a nuisance as far as debugging and similar goes.

As of now, having something in the deploy directory is pretty much as painful as having it in code, though.

controllers/application/configmap.go
anderssonw: from #152 i've noticed the error handling is somewhat different between the two configmaps, so i'm not sure if it would make much sense to create a general function

pkg/util/helperfunctions.go
esphen: I agree, it's not ideal. If we put it in deploy new deploys will overwrite the changes you've put on the cluster, if it doesn't get terraform state mismatch issues which is worse somehow. On the other hand config drift is a huge issue and cause of bugs if not handled in a good way.

I guess I would put them in deploy for now, then improve this when we migrate deploy to argo. But leaving them out is also an option, but then we need some kind of process to avoid config drift. Your call

controllers/application/configmap.go
esphen: Does the error handling need to be different?

pkg/util/helperfunctions.go
anderssonw: @evenh @omaen would love to hear y'alls' opinions :)

how i see it we have the following options:
1. Hard-coded namespaces with an additional configmap used for PoCs and stuff. This is how the PR looks now.
2. Exclusively a ConfigMap generated by Skiperator which we can add/delete to as we please. This is the most agile, but differences between environments is 100% going to happen.
3. Deployed ConfigMap instead of hard-coded list, with the ability to edit. Will make a sort of "single source of truth", but issues with state might occur if we append to the ConfigMap manually.

evenh: I do agree with @omaen's remarks.

controllers/application/configmap.go
evenh: If deciding on the more generic `GetConfigMap` approach, one could add functions for error handling like `util.ErrStopReconciling` and `util.ErrDoPanic`

anderssonw:
@esphen @evenh @omaen 

Changes according to comments have been made :)

controllers/namespace/controller.go
```diff
@@ -19,15 +22,62 @@ type NamespaceReconciler struct {
...
+	namespaceExclusionCMap, err := util.GetConfigMap(r.GetClient(), ctx, configMapNamespacedName)
+	if err != nil {
+		util.ErrDoPanic(err, "Something went wrong getting namespace-exclusion config map: %v")
+	}
+
+	nameSpacesToExclude := namespaceExclusionCMap.Data
+
+	exclusion, keyExists := nameSpacesToExclude[namespace]
+
+	return (keyExists && exclusion == "true") || slices.Contains(excludedNamespaceList, namespace)
```
esphen, line 71: I'm not sure if this is the final version or if you're waiting for the discussion above on what to use as the source of truth.

I was expecting the `excludedNamespaceList` in code to be replaced by the deployed configmap. Is this for backwards compatibility? Do you mind explaining why we need both?

pkg/util/helperfunctions.go
esphen: 3 looks the most appealing. Issues with mismatching terraform state can be solved by migrating deployment to Argo which properly diffs desired state against actual state instead of a tertiary source of truth.

controllers/namespace/controller.go
anderssonw: i simply forgot removing the in-code list. update coming

esphen (approved):

controllers/namespace/controller.go
```diff
@@ -19,15 +21,28 @@ type NamespaceReconciler struct {
...
+		util.ErrDoPanic(err, "Something went wrong getting namespace-exclusion config map: %v")
+	}
+
+	nameSpacesToExclude := namespaceExclusionCMap.Data
+
+	exclusion, keyExists := nameSpacesToExclude[namespace]
+
+	return (keyExists && exclusion == "true")
+}
+
```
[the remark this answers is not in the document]
anderssonw: In this case its logging that something, which causes a panic, went wrong. Panic halts execution doesn't it? So additional logging probably won't be a huge issue.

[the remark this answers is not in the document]
anderssonw: aaaaaah, my bad, misunderstood :)

PR #152, opened 2023-02-07
[SKIP-958] Skiperator Config Map adds Pod request for istio proxies

anderssonw:
This PR adds a new flag during Skiperator startup which sets the environment in which Skiperator is started. Based on this environment, an annotation is added to deployment pods for changing their proxy sidecar CPU requests. For sandbox/dev/test, this is set to 10m, and for production its set to the usual 100m.

Tests have been added to make sure the template spec is correct.

Worth noting: The deployment spec is not updated upon changing the config map, and as such the deployment needs to be reconciled somehow else. This seems odd, as data changes to config maps is mostly how you would change a config map. So if the ConfigMap is changed, a reconcile must be forced somehow. Not a huge fan of this but not sure how to circumvent.


controllers/application/configmap.go
```diff
@@ -99,3 +127,56 @@ func (r *ApplicationReconciler) setupGCPAuthConfigMap(ctx context.Context, gcpId
...
+	if len(mapData) == 0 {
+		mapData = make(map[string]string)
+	}
+
+	// // We only want to set the istio CPU Request if it is not already set
+	// _, present := mapData[istioSidecarAnnotation]
+	// if !present {
+	// mapData[istioSidecarAnnotation] = getDefaultIstioCPURequestFromEnv(r.Environment)
+	// }
+	// ^ Do we only want to do this if its not already set? Should we allow overriding the default request per application?
```
anderssonw, line 149: Would like some comments here. You cannot edit the ConfigMap now.

(Well you can, and it wont be reconciled, but as soon as a deployment is reconciled, so is the configmap and your changes will not matter)

evenh: A common pattern is to add a suffix to each revision of the configmap to trigger a redeployment. E.g.

```
foo-skiperator-config-ffaa00
```

where `ffaa00` is a hash over the ConfigMap's contents. This specific name is in turn referenced in the `skiperator` deployment. When content in the ConfigMap changes, a new name is generated and the deployment is updated to use the new ConfigMap.

---

I'm not sure if that's the best option here or not – let's discuss offline

controllers/application/configmap.go
```diff
@@ -99,3 +127,56 @@ func (r *ApplicationReconciler) setupGCPAuthConfigMap(ctx context.Context, gcpId
...
+func getDefaultIstioCPURequestFromEnv(env string) string {
+	switch env {
+	case "prod":
+		return "100m"
+	case "sandbox", "dev", "test":
+		return "10m"
+	default:
+		// Better to safeguard a high request in case of poor config
+		// Set to something other than production just to tell difference
+		return "99m"
```
evenh, line 175: That's a very useful default 👏🏻

PR #156, opened 2023-02-10
ArgoCd external URL

EbiJon:
We have added an additional external links to ArgoCD dashboard


skiperator/controllers/application/deployment.go
```diff
@@ -0,0 +1,232 @@
+package applicationcontroller
```
evenh, line 1: Is this a local backup that was committed in by accident?

controllers/application/deployment.go
```diff
@@ -13,6 +13,8 @@ import (
 	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
+// Adding an argocd external link constant
```
evenh, line 16:
```suggestion
```

controllers/application/deployment.go
```diff
@@ -13,6 +13,8 @@ import (
...
+const ( AnnotationKeyLinkPrefix = "link.argocd.argoproj.io/external-link" )
```
evenh, line 17:
```suggestion
const (
	AnnotationKeyLinkPrefix = "link.argocd.argoproj.io/external-link"
)
```

or

```suggestion
const AnnotationKeyLinkPrefix = "link.argocd.argoproj.io/external-link"
```

See also `go fmt` which automatically formats code according to the [accepted format](https://go.dev/doc/effective_go#formatting).

controllers/application/deployment.go
```diff
@@ -44,6 +46,12 @@ func (r *ApplicationReconciler) reconcileDeployment(ctx context.Context, applica
 		}
 
 		r.SetLabelsFromApplication(ctx, &deployment, *application)
+		
+		// add an external link to argocd
+		ingresses := application.Spec.Ingresses
+		if len(ingresses) > 0 {
+			deployment.ObjectMeta.Annotations[AnnotationKeyLinkPrefix] =  ingresses[0]
```
evenh, line 53: What do we do when there are multiple ingresses specified? 

Open question: Is `Deployment` the correct resource to place the annotation at?

evenh (changes requested):
A few comments. Would also like to see some tests to verify the introduced annotation.

controllers/application/deployment.go
esphen: For reference: https://argo-cd.readthedocs.io/en/stable/user-guide/external-url/

> What do we do when there are multiple ingresses specified?

The annotation only supports one URL per resource, so my suggestion was to take the first URL and ignore the others. Then we can document that the first URL will be used for links in the UI.

> Open question: Is Deployment the correct resource to place the annotation at?

I assumed only deployments are supported from the example in the documentation above, but reading the text closer it doesn't specify that as a limitation. We could consider putting it on another resource.

skiperator/controllers/application/deployment.go
esphen: Agreed, it seems this file was added by accident and should be removed before merging. It's still in the branch so I'm unresolving this thread

controllers/application/deployment.go
evenh: Let's start with one URL per application, at the `Deployment` level and iterate if we see other requirements emerge.

EbiJon (commented):
Edit constant

controllers/application/deployment.go
```diff
@@ -44,6 +48,13 @@ func (r *ApplicationReconciler) reconcileDeployment(ctx context.Context, applica
 		}
 
 		r.SetLabelsFromApplication(ctx, &deployment, *application)
+		
+		// add an external link to argocd
+		ingresses := application.Spec.Ingresses
+		if len(ingresses) > 0 {
+			deployment.ObjectMeta.Annotations = make(map[string]string, 1)
+			deployment.ObjectMeta.Annotations[AnnotationKeyLinkPrefix] = fmt.Sprintf("https://%s", ingresses[0])
+		}
```
anderssonw, line 57: This will overwrite any other Annotations a deployment has set. I'm not sure if argo or any other framework in our cluster adds annotations, but it could be preferable to avoid overwriting just in case. We had this issue with labels earlier, so it could show up.

tests/argocd_external_link/00-assert.yaml
```diff
@@ -0,0 +1,21 @@
...
+  template:
+    metadata:
+      annotations:
+        prometheus.io/scrape: "true"
+      labels:
+        app: argocd-external-link
+    spec:
+      containers:
+        - name: argocd-external-link
+          image: image
```
anderssonw, line 21: when asserting using ´kuttl´ you only need to check for what you are adding. in this case the spec of the Deployment is not really interesting, but whether or not the annotation for argo is. i would remove the spec entirely.

evenh (approved):

controllers/application/deployment.go
```diff
@@ -52,6 +58,12 @@ func (r *ApplicationReconciler) reconcileDeployment(ctx context.Context, applica
 			"prometheus.io/scrape":            "true",
 		}
 
+		// add an external link to argocd
+		ingresses := application.Spec.Ingresses
+		if len(ingresses) > 0 {
+			deployment.ObjectMeta.Annotations[AnnotationKeyLinkPrefix] = fmt.Sprintf("https://%s", ingresses[0])
```
anderssonw, line 64: This will fail if the Annotations map is a nil map. I'm not sure this will happen, but its worth having a check for it. Check out how its done with Application labels here: https://github.com/kartverket/skiperator/blob/main/pkg/util/reconciler.go#L143

anderssonw (changes requested):

anderssonw:
Somehow the tests work just fine now though, thats odd 🤔

controllers/application/deployment.go
evenh: It's explicitly set on line 56, so it can't be nil. That's because this PR was updated against `main`, which introduces the argocd annotation.

anderssonw: aha, i missed that :)

anderssonw (approved):

PR #160, opened 2023-02-14
Stop argo pruning skiperator owned resources

esphen:
When Argo and Skiperator run in the same system they conflict when tearing down a namespace. This PR tells Argo not to touch skiperator app generated resources so namespaces don't get into an infinite terminating state.

Uses this annotation: https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#no-prune-resources

Tested in sandbox. Seems to work.

### DIFF
--- a/controllers/application/deployment.go
+++ b/controllers/application/deployment.go
@@ -44,7 +44,13 @@ func (r *ApplicationReconciler) reconcileDeployment(ctx context.Context, applica
 
 		r.SetLabelsFromApplication(ctx, &deployment, *application)
 
-		deployment.Spec.Template.ObjectMeta.Annotations = map[string]string{"prometheus.io/scrape": "true"}
+		deployment.ObjectMeta.Annotations = map[string]string{
+			"argocd.argoproj.io/sync-options": "Prune=false",
+		}
+		deployment.Spec.Template.ObjectMeta.Annotations = map[string]string{
+			"argocd.argoproj.io/sync-options": "Prune=false",
+			"prometheus.io/scrape": "true",
+		}
 
 		labels := map[string]string{"app": application.Name}
 		deployment.Spec.Template.ObjectMeta.Labels = labels
--- a/controllers/application/deployment.go
+++ b/controllers/application/deployment.go
@@ -2,7 +2,9 @@ package applicationcontroller
 
 import (
 	"context"
+
 	skiperatorv1alpha1 "github.com/kartverket/skiperator/api/v1alpha1"
+	util "github.com/kartverket/skiperator/pkg/util"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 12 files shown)


controllers/application/deployment.go
```diff
@@ -44,7 +46,11 @@ func (r *ApplicationReconciler) reconcileDeployment(ctx context.Context, applica
 
 		r.SetLabelsFromApplication(ctx, &deployment, *application)
 
-		deployment.Spec.Template.ObjectMeta.Annotations = map[string]string{"prometheus.io/scrape": "true"}
+		deployment.ObjectMeta.Annotations = util.CommonAnnotations
+		deployment.Spec.Template.ObjectMeta.Annotations = map[string]string{
+			"argocd.argoproj.io/sync-options": "Prune=false",
+			"prometheus.io/scrape":            "true",
+		}
```
evenh, line 53: Since this might be common pattern going forward (common annotation + a special case) I propose the following
```suggestion
		deployment.Spec.Template.ObjectMeta.Annotations = util.CommonAnnotationsPlus(
			"prometheus.io/scrape", "true",
		)
```

And adding the following function in `pkg/util/helperfunctions.go`:
```go
func CommonAnnotationsPlus(kvs ...string) map[string]string {
	if len(kvs)%2 != 0 {
		panic(fmt.Sprintf("both key and value must be supplied, got %s", kvs))
	}

	m := CommonAnnotations
	for i := 0; i < len(kvs); i = i + 2 {
		m[kvs[i]] = kvs[i+1]
	}

	return m
}
```

evenh (approved):
LGTM, added a suggestion but feel free to ignore :-)

esphen:
> Is it possible to do it the other way around and set a Skiperator specific-annotation on these resources, and tell Argo to ignore them? Just seems a bit intrusive from Skiperators point of view to have to set this annotation.

Not sure exactly what you mean, the annotation that is set here makes Argo ignore these resources. Otherwise Argo will try to delete them when terminating a namespace which causes a deadlock when skiperator's finalizers try to delete them again. It's pretty much the same as setting the `prometheus/scrape` annotation on behalf of the teams as we already do.

It would probably be even more intrusive if we were to do it the other way around and add some feature that made it possible for skiperator to create, but not delete resources so that argo would do deletions. Right now that is done implicitly through the ownership pattern, which tells skiperator what it owns so that it automatically cleans it up afterwards.

esphen:
> Just wanted to check if one could use an annotation such as `skiperator.kartverket.no/managed: true` on the resources that should be owned by skiperator, and in some way tell Argo to skip them

Unfortunately not, it has to be this specific annotation.

>  Maybe add a note about the annotation in the contributing doc?

Good idea, I'll document this property before merging

PR #161, opened 2023-02-17
[SKIP-983] Add common annotation util

anderssonw:
Annotations added automatically, in this case `"deployment.kubernetes.io/revision": "7"`, were overwritten by Skiperator, forcing looping reconciles.

Added a helper util for merging the current annotations with the common annotations.

### DIFF
--- a/controllers/application/horizontal_pod_autoscaler.go
+++ b/controllers/application/horizontal_pod_autoscaler.go
@@ -4,7 +4,7 @@ import (
 	"context"
 
 	skiperatorv1alpha1 "github.com/kartverket/skiperator/api/v1alpha1"
-	util "github.com/kartverket/skiperator/pkg/util"
+	"github.com/kartverket/skiperator/pkg/util"
 	autoscalingv2 "k8s.io/api/autoscaling/v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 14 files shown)


pkg/util/helperfunctions.go
```diff
@@ -58,3 +61,12 @@ func GenerateHashFromName(name string) uint64 {
 	_, _ = hash.Write([]byte(name))
 	return hash.Sum64()
 }
+
+func SetCommonAnnotations(context context.Context, object client.Object) {
```
evenh, line 65: Unused context - future proofing or an error?

controllers/application/horizontal_pod_autoscaler.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 
 	skiperatorv1alpha1 "github.com/kartverket/skiperator/api/v1alpha1"
-	"github.com/kartverket/skiperator/pkg/util"
+	util "github.com/kartverket/skiperator/pkg/util"
```
evenh, line 7: This package alias is redundant (every import)

pkg/util/helperfunctions.go
anderssonw: Error, used the context in the helper function at first.

evenh (approved):

esphen (approved):

PR #187, opened 2023-03-29
[SKIP-1004] HTTPS redirect for ingresses

anderssonw:
Sets up a new Virtual Service for redirecting `port: 80` without URI `/.well-known/acme-challenge` to HTTPS.

A new field in the CRD, `RedirectToHTTPS`, manages whether or not this Virtual Service is created.

Added the creation of this VirtualService and its deletion upon switching off `RedirectToHTTPS` to the ingress tests.

Removes predicate in controller setup for VirtualServices. Only Owned VirtualServices are reconciled upon, so no need for this.

Also attempted to use a more struct based initialisation of Spec for VirtualServices.

Also added a deletion step to all tests using Application. Tests were running slow and this improved the time spent by a big margin.


controllers/application/ingress_gateway.go
```diff
@@ -37,30 +37,42 @@ func (r *ApplicationReconciler) reconcileIngressGateway(ctx context.Context, app
 			r.SetLabelsFromApplication(ctx, &gateway, *application)
 			util.SetCommonAnnotations(&gateway)
 
+			internalSelector := "istio-ingress-internal"
+			externalSelector := "istio-ingress-external"
```
evenh, line 41:
```suggestion
			internalSelector := map[string]string{"app": "istio-ingress-internal"}
			externalSelector := map[string]string{"app": "istio-ingress-external"}
```

controllers/application/ingress_gateway.go
```diff
@@ -37,30 +37,42 @@ func (r *ApplicationReconciler) reconcileIngressGateway(ctx context.Context, app
...
+
 			if util.IsInternal(hostname) {
-				gateway.Spec.Selector = map[string]string{"app": "istio-ingress-internal"}
+				gateway.Spec.Selector = map[string]string{"app": internalSelector}
```
evenh, line 44:
```suggestion
				gateway.Spec.Selector = internalSelector
```

controllers/application/ingress_gateway.go
```diff
@@ -37,30 +37,42 @@ func (r *ApplicationReconciler) reconcileIngressGateway(ctx context.Context, app
...
 			} else {
-				gateway.Spec.Selector = map[string]string{"app": "istio-ingress-external"}
+				gateway.Spec.Selector = map[string]string{"app": externalSelector}
```
evenh, line 46:
```suggestion
				gateway.Spec.Selector = externalSelector
```

controllers/application/ingress_virtual_service.go
```diff
@@ -71,8 +73,120 @@ func (r *ApplicationReconciler) reconcileIngressVirtualService(ctx context.Conte
...
+										Prefix: "/.well-known/acme-challenge/",
+									},
+								},
+							},
+							Port: 80,
+						},
+					},
+					Redirect: &networkingv1beta1api.HTTPRedirect{
+						Scheme:       "https",
+						RedirectCode: 302,
```
evenh, line 102:
```suggestion
						RedirectCode: 308,
```

> Although the [specification](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301#specifications) requires the method and the body to remain unchanged when the redirection is performed, not all user-agents meet this requirement. Use the 301 code only as a response for [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) or [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD) methods and use the [308 Permanent Redirect](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308) for [POST](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST) methods instead, as the method change is explicitly prohibited with this status.

See MDN on [301](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301) and [308](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308).

controllers/application/ingress_virtual_service.go
```diff
@@ -71,8 +73,120 @@ func (r *ApplicationReconciler) reconcileIngressVirtualService(ctx context.Conte
...
+		if err != nil {
+			r.SetControllerError(ctx, &application, controllerName, err)
+			return reconcile.Result{}, err
+		}
+	}
+
+	return reconcile.Result{}, err
+}
+
+var EXPORT_TO_NAMESPACES = []string{".", "istio-system", "istio-gateways"}
```
evenh, line 192:
```suggestion
var ExportToNamespaces = []string{".", "istio-system", "istio-gateways"}
```

Use snakeCase and move declaration to the top of the file, under imports.

api/v1alpha1/application.go
```diff
@@ -65,6 +65,11 @@ type ApplicationSpec struct {
 
 	//+kubebuilder:validation:Optional
 	Ingresses []string `json:"ingresses,omitempty"`
+
+	//+kubebuilder:validation:Optional
+	//+kubebuilder:default=false
```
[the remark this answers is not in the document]
anderssonw: Not in particular. Makes more sense to default it to true, but didn't want any unwanted side effects for product teams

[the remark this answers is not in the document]
evenh: I see. Let's have a goal to switch the default value in 30-60 days when we have had the opportunity to talk with product teams :-)

evenh (approved):

PR #202, opened 2023-04-20
Add support for Pod Disruption Budgets

evenh:
In order to reduce possible disruptions and increase availability, support for PDBs is introduced in this PR. The logic is pretty simple:

If number of `minReplicas` is greater than 1, specify a PDB with `minAvailable` set to `50%`, otherwise `minAvailable` is set to `0`.


esphen:
Heier på denne! 🎉

anderssonw (approved):

PR #203, opened 2023-04-20
Add revisionhistorylimit to deployment

anderssonw:
Every skiperator application now gets 10 replicasets, which is quite annoying when viewing in ArgoCD: 
image

Changing this to 2 should make it nicer, while still keeping some RSs


evenh (approved):
Nice improvement, reducing signal-to-noise ratio 👏

PR #204, opened 2023-04-20
Change Skiperator to prefer Struct init and modularization of functions

anderssonw:
Skiperator resource definitions often do a lot, especially for the deployment resource. This PR aims to prefer the use of Structs for initing resources, as well as handling most of the definitions in their own functions where extra logic is required.


esphen:
Useful and more legible, but please test this out first and report back your results. My experience using structs in the controllers like this has been infinite reconcile loops, most likely due to instantiating new objects that are not referentially equal to the previous state. If we see a spike in the amount of reconciliation metrics after deploying this we should take another look at what we can do to stop that. Otherwise I'm in support of this.

anderssonw:
@evenh @esphen wanna have a 👀 ?

evenh (approved):
Agree with the previous comments regarding reconcile performance. Overall a very welcome change.

PR #206, opened 2023-04-21
[SKIP-956] Versioning for Skiperator

anderssonw:
Uses Goreleaser to create a versioning scheme on new tags for Skiperator

See https://github.com/kartverket/skiperator/releases/tag/v1.0.0-alpha for a release made by the workflow upon tag creation


esphen:
Cool!
Will this change the format of the tags in such a way that it might affect other places we use the tags?
Is it trivial to extend this with [cosign signing](https://goreleaser.com/customization/sign/?h=sbom#signing-with-cosign) and SBOM at a later stage?

anderssonw:
> Cool! Will this change the format of the tags in such a way that it might affect other places we use the tags? Is it trivial to extend this with [cosign signing](https://goreleaser.com/customization/sign/?h=sbom#signing-with-cosign) and SBOM at a later stage?

Cosign seems to operate well with image tags, as well as the policy controller being able to spot when a tag has a new image.

I'm not entirely sure about the built-in cosign sign, but will look into it.

Also not entirely clear how we should change other parts of our deployment if we start to use the releases created here. For the time being this is mostly to have some versioning, which is useful for both godocs (see how out of date https://pkg.go.dev/github.com/kartverket/skiperator/cmd/skiperator is) and proper changelogs. Tags over digests in deployments or vice versa is a discussion we also need to have :)

evenh:
This seems like a nice resource: https://github.com/goreleaser/goreleaser-example-supply-chain

.goreleaser.yaml
evenh, line 1: Future improvement: use another OIDC issuer

evenh (approved):

BardOve (approved):

PR #207, opened 2023-04-21
[SKIP-616] Move SeccompProfile from Container to Pod level

esphen:
Suggested fix to allow for using [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission) in enforce mode alongside Istio sidecars.

Moves SeccompProfile from container level to pod level so it applies to the istio sidecars as well as the application container, which should make PSA happy.

See: https://github.com/istio/istio/issues/35894


evenh (changes requested):
`resource Deployment:test/minimal: .spec.template.spec.containers.securityContext.seccompProfile: key is missing from map`

esphen:
@evenh Tester oppdatert, da er de grønne også

evenh (approved):

PR #213, opened 2023-05-05
SKIP-1052 Validate Ingresses and Redirect https default true

anderssonw:
Created a wrapper for Validation Functions and a PoC for how to validate different resources with proper errors. Not sure if this is how we should handle validation long term, but it seems to suffice now.

Ingresses can not be empty, contains spaces or uppercase letters.

Result of a malformed ingress here:
image

RedirectToHTTPS defaults to true.

### DIFF
--- a/api/v1alpha1/application.go
+++ b/api/v1alpha1/application.go
@@ -72,8 +72,8 @@ type ApplicationSpec struct {
 	Ingresses []string `json:"ingresses,omitempty"`
 
 	//+kubebuilder:validation:Optional
-	//+kubebuilder:default=true
-	RedirectToHTTPS bool `json:"redirectToHTTPS,omitempty"`
+	//+kubebuilder:default:=true
+	RedirectToHTTPS *bool `json:"redirectToHTTPS,omitempty"`
 
 	//+kubebuilder:validation:Optional
 	AccessPolicy AccessPolicy `json:"accessPolicy,omitempty"`
--- a/api/v1alpha1/application.go
+++ b/api/v1alpha1/application.go
@@ -66,11 +66,14 @@ type ApplicationSpec struct {
 	//+kubebuilder:validation:Optional
 	Startup *Probe `json:"startup,omitempty"`
 
-	// Ingresses must be lowercase, contain no spaces, and be non-empty
+	// Ingresses must be lower case, contain no spaces, be a non-empty string, and have a hostname/domain separated by a period
 	//
 	//+kubebuilder:validation:Optional
 	Ingresses []string `json:"ingresses,omitempty"`
 
+	// Controls whether or not the application will automatically redirect all HTTP calls to HTTPS via the istio VirtualService.
+	// This redirect does not happen on the route /.well-known/acme-challenge/, as the ACME challenge can only be done on port 80.
+	//
 	//+kubebuilder:validation:Optional
 	//+kubebuilder:default:=true
 	RedirectToHTTPS *bool `json:"redirectToHTTPS,omitempty"`
--- a/api/v1alpha1/application.go
+++ b/api/v1alpha1/application.go
@@ -66,11 +66,13 @@ type ApplicationSpec struct {
 	//+kubebuilder:validation:Optional
 	Startup *Probe `json:"startup,omitempty"`
 
+	// Ingresses must be lowercase, contain no spaces, and be non-empty
+	//
 	//+kubebuilder:validation:Optional
 	Ingresses []string `json:"ingresses,omitempty"`
 
 	//+kubebuilder:validation:Optional
-	//+kubebuilder:default=false
+	//+kubebuilder:default=true
 	RedirectToHTTPS bool `json:"redirectToHTTPS,omitempty"`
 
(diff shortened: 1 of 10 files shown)


controllers/application/controller.go
```diff
@@ -207,3 +220,32 @@ func (r *ApplicationReconciler) finalizeApplication(ctx context.Context, applica
...
+
+	return nil
+}
+
+func ValidateIngresses(application *skiperatorv1alpha1.Application) error {
+	for _, ingress := range application.Spec.Ingresses {
+		if ingress == "" || util.HasUpperCaseLetter(ingress) || strings.Contains(ingress, " ") {
+			errMessage := fmt.Sprintf("ingress with value '%s' was not valid. ingress must be lower case, contain no spaces, and be a non-empty string", ingress)
+			return errors.NewInvalid(application.GroupVersionKind().GroupKind(), application.Name, field.ErrorList{
+				field.Invalid(field.NewPath("application").Child("spec").Child("ingresses"), application.Spec.Ingresses, errMessage),
```
anderssonw, line 245: this section can probably be generalised somewhat

pkg/util/helperfunctions.go
```diff
@@ -71,3 +72,15 @@ func PointTo[T any](x T) *T {
...
+	hasUpper := false
+	for _, letter := range word {
+		if unicode.IsUpper(letter) {
+			hasUpper = true
+			break
+		}
+	}
+
+	return hasUpper
+}
```
evenh, line 86:
```suggestion
func HasUpperCaseLetter(word string) bool {
	for _, letter := range word {
		if unicode.IsUpper(letter) {
			return true
		}
	}

	return false
}
```

api/v1alpha1/application.go
```diff
@@ -66,11 +66,13 @@ type ApplicationSpec struct {
 	//+kubebuilder:validation:Optional
 	Startup *Probe `json:"startup,omitempty"`
 
+	// Ingresses must be lowercase, contain no spaces, and be non-empty
```
esphen, line 69: Add this to the README as well?

pkg/util/helperfunctions.go
```diff
@@ -71,3 +72,15 @@ func PointTo[T any](x T) *T {
 func GetApplicationSelector(applicationName string) map[string]string {
 	return map[string]string{"app": applicationName}
 }
+
+func HasUpperCaseLetter(word string) bool {
+	hasUpper := false
+	for _, letter := range word {
+		if unicode.IsUpper(letter) {
+			hasUpper = true
+			break
```
esphen, line 81:
```suggestion
			return true
```

Would an early return here work better?

controllers/application/controller.go
```diff
@@ -91,6 +94,16 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req reconcile.Req
 		return reconcile.Result{}, err
 	}
 
+	err = r.validateApplicationSpec(application)
+	if err != nil {
+		r.GetRecorder().Eventf(
+			application,
+			corev1.EventTypeWarning, "InvalidApplication",
+			"Application was not valid, error: %s", err.Error(),
```
esphen, line 102:
```suggestion
			"Application failed validation and was rejected, error: %s", err.Error(),
```
How do you feel about this text?

esphen (commented):
Looks promising!

controllers/application/controller.go
anderssonw: better wording indeed :)

controllers/application/controller.go
anderssonw: definitely possible. i'm down to change it to a regex instead of multiple if conditions if that's preferable.

it *is* more readable this way, though.

api/v1alpha1/application.go
```diff
@@ -66,12 +66,14 @@ type ApplicationSpec struct {
...
+	//
 	//+kubebuilder:validation:Optional
 	Ingresses []string `json:"ingresses,omitempty"`
 
 	//+kubebuilder:validation:Optional
-	//+kubebuilder:default=false
-	RedirectToHTTPS bool `json:"redirectToHTTPS,omitempty"`
+	//+kubebuilder:default:=true
+	RedirectToHTTPS *bool `json:"redirectToHTTPS,omitempty"`
```
esphen, line 76: Trenger denne å kunne være nil?

controllers/application/controller.go
```diff
@@ -207,3 +220,32 @@ func (r *ApplicationReconciler) finalizeApplication(ctx context.Context, applica
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func ValidateIngresses(application *skiperatorv1alpha1.Application) error {
+	for _, ingress := range application.Spec.Ingresses {
```
esphen, line 241: Hva skjer om `application.Spec.Ingresses` er nil (udefinert)? Funker det å range over da?

anderssonw: would give 0 iterations

esphen (approved):

PR #220, opened 2023-05-19
Add support for Prometheus Operator CRD (ServiceMonitor)

evenh:
SKIP-1073


anderssonw (approved):

esphen:
@evenh Would it be too heavy to query for the existence of the CRD every reconcile? If not then that would probably be a good approach, in the sense that we don't need to make any API changes.

evenh:
Will look into it.

berg-thom (approved):

PR #225, opened 2023-05-22
K8s 1.27

evenh:
This PR attempts to address the different incompatibilites from the dependabot PRs regarding K8s 1.27. I had to remove the `kubectl kuttl` plugin in order to make the tests pass locally. As far as I know, that's not something that's in use.


evenh:
`controller-runtime` don't follow semver, and there were API breakage introduced between `v0.14.6` and `v0.15.0-beta.0` (used in this PR)

anderssonw:
> `controller-runtime` don't follow semver, and there were API breakage introduced between `v0.14.6` and `v0.15.0-beta.0` (used in this PR)

aha, that's annoying..

PR #226, opened 2023-05-24
Documentation in application.go over README

anderssonw:
We should probably have a way of releasing Skiperator before this is merged (see #206), as the resulting godocs should be available at https://pkg.go.dev/github.com/kartverket/skiperator/api/v1alpha1, where it would be easier to read than inline comments in application.go.

We should even be able to add code examples where there might be more confusion, such as proper examples for AccessPolicy


esphen (approved):
I love this, autogenerated docs is the way to go 🙂 

The only thing which I think will be kind of annoying is that pkg.go.dev seems to include the kubebuilder commands as a normal doc string, thus we will see them in the docs page. That can be distracting. Don't know if there is a way to get rid of that

lislei (approved):

anderssonw:
> I love this, autogenerated docs is the way to go 🙂
> 
> The only thing which I think will be kind of annoying is that pkg.go.dev seems to include the kubebuilder commands as a normal doc string, thus we will see them in the docs page. That can be distracting. Don't know if there is a way to get rid of that

I agree this is annoying.. Presumably there has to be some way to ignore it? I will ask around :)

Might have to do some extra doc generation somehow, but that may be worth it

PR #233, opened 2023-06-05
Expose Prometheus metrics at :8181

evenh:
Metrics are gathered but never exposed.


cmd/skiperator/main.go
```diff
@@ -79,6 +79,7 @@ func main() {
 		HealthProbeBindAddress:  ":8081",
 		LeaderElection:          *leaderElection,
 		LeaderElectionNamespace: *leaderElectionNamespace,
+		MetricsBindAddress:      ":8181",
```
esphen, line 82: Må det settes opp service for dette også?

esphen (approved):

PR #238, opened 2023-06-13
Add typemeta to deployment init

anderssonw:
If typemeta is not defined on deployment struct init, resource labels are not added due to the source object not having its typemeta set before the actual creation of the resource.

This caused gatekeeper to not allow replicalimits to be less than 2 even if the resource label for excluding this rule was added.


evenh (approved):
Nice catch 👏🏻

PR #239, opened 2023-06-15
Add app label to service

esphen:
When using ServiceMonitors, it creates a selector that selects a service. This currently uses the `app` label, which is not present on the service (in fact there are no labels on it at all). This PR adds labels to the service.

### DIFF
--- a/controllers/application/service.go
+++ b/controllers/application/service.go
@@ -28,6 +28,11 @@ func (r *ApplicationReconciler) reconcileService(ctx context.Context, applicatio
 		r.SetLabelsFromApplication(ctx, &service, *application)
 		util.SetCommonAnnotations(&service)
 
+		// ServiceMonitor requires labels to be set on service to select it
+		labels := service.GetLabels()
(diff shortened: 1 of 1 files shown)


controllers/application/service.go
```diff
@@ -28,6 +28,11 @@ func (r *ApplicationReconciler) reconcileService(ctx context.Context, applicatio
 		r.SetLabelsFromApplication(ctx, &service, *application)
 		util.SetCommonAnnotations(&service)
 
+		// ServiceMonitor requires labels to be set on service to select it
+		labels := service.GetLabels()
+		labels["app"] = application.Name
+		service.SetLabels(labels)
```
anderssonw, line 34: `util.GetApplicationSelector` returnerer egentlig det du lager der. Spesielt om servicen ikke har noen labels fra før. finnes kanskje en map merge funksjon eller liknende?

anderssonw (approved):

evenh (approved):

berg-thom (approved):

PR #247, opened 2023-06-21
Add flag for toggling PDBs

evenh:
Adds an optional flag to disable PodDisruptionBudget creation.


BardOve (approved):

api/v1alpha1/application.go
```diff
@@ -78,6 +78,12 @@ type ApplicationSpec struct {
 	//+kubebuilder:default:=true
 	RedirectToHTTPS *bool `json:"redirectToHTTPS,omitempty"`
 
+	// Whether to enable automatic Pod Disruption Budget creation for this application.
+	//
+	//+kubebuilder:validation:Optional
+	//+kubebuilder:default=true
+	EnablePDB bool `json:"enablePDB"`
```
anderssonw, line 85: If you set the EnablePDB to a *bool and add the `omitempty` flag for the json serializer, the field will be unset if it is actually unset. I guess the default will make it never be omitted, but I think in general its preferable to have it omittable :)

anderssonw (approved):
lgtm, one small comment but that might just be personal preference

PR #252, opened 2023-06-26
Rework Prometheus metrics scraping

evenh:
This PR adds a new section to the CRD:
```yaml
# Optional settings for how Prometheus compatible metrics should be scraped.
prometheus:
  port: 8181         # *Required* Pod port number or name
  path: "/metrics"   # *Optional* Defaults to /metrics
```

If specified, it will create a matching `ServiceMonitor` and tweak the `NetworkPolicy` to accept scrape traffic from grafana-agent.


controllers/application/service_monitor.go
```diff
@@ -21,12 +19,16 @@ func (r *ApplicationReconciler) reconcileServiceMonitor(ctx context.Context, app
 		return reconcile.Result{}, nil
 	}
 
-	if !hasMetricsPort(application) {
+	if application.Spec.Prometheus == nil {
 		r.SetControllerFinishedOutcome(ctx, application, controllerName, nil)
 		return reconcile.Result{}, nil
 	}
```
anderssonw, line 25: i believe you should add a client.Delete() for the service monitor here. I assume that removing the prometheus config will not remove the resource itself as of now

anderssonw (changes requested):
nice! one small thing regarding deletion of service monitor :)

controllers/application/service_monitor.go
evenh: Good catch!

anderssonw (approved):
smoothie mcflip

BardOve (approved):

PR #261, opened 2023-07-09
Allow port number or name in probes

evenh:
Removing a self-imposed limitation.

See the updated [HTTPGetAction in the v1 k8n spec](https://github.com/kubernetes/api/blob/master/core/v1/types.go)

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ spec:
   liveness:
     # The path to access on the HTTP server
     path: /healthz
-    # Number of the port to access on the container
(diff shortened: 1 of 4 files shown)


lislei (approved):
:shipit: 
In stead of port number, you now have the option to use an `IANA_SVC_NAME` type of name for binding your service ports in k8n. 

- The schema-changes seem backwards compatible to me.
- The implementation improvements are reflected in the schema elements for `liveness`, `readiness` and `startup`.

BardOve (approved):

PR #267, opened 2023-07-14
Only diff a cleaned up deployment against the definition

evenh:
This resolves a warning with Instana. I'm happy to hear about any better approaches, but this one seems to work in dev.


esphen:
Er dette pga. vi har issues med instana autoinstrumentation? Bruker vi det sånn egentlig? Skulle vi skrudd det av?

anderssonw:
@esphen I spoke quickly about this with @evenh 

We both agree that, in the case that we can completely ignore instana and turn it off, that would be a preferable solution. We have a few workarounds going on here, but they are all somewhat annoying and we would rather just.. not.

anderssonw:
Alright to close this? @evenh

evenh:
Please do!

PR #272, opened 2023-07-24
[SKIP-705] Extract resource generation to own packages

anderssonw:
Extracts some functions which generate resources to its own library. Avoids having the future SKIPJob controller fetch from the Application controller package, instead creating a new one. Only done for resources which will be used by SKIPJobs in the future, see #227

Not sure where exactly to put the istio constants and the `IstioEnabled()` function. If there are any inputs as to where this should be placed let me know :)

### DIFF
--- a/controllers/application/network_policy.go
+++ b/controllers/application/network_policy.go
@@ -22,15 +22,17 @@ func (r *ApplicationReconciler) reconcileNetworkPolicy(ctx context.Context, appl
 		return reconcile.Result{}, err
 	}
 
-	netpolSpec := networking.CreateNetPolSpec(networking.NetPolOpts{
-		AccessPolicy:     application.Spec.AccessPolicy,
-		Ingresses:        &application.Spec.Ingresses,
-		Port:             &application.Spec.Port,
-		Namespace:        application.Namespace,
-		Name:             application.Name,
-		RelatedServices:  &egressServices,
-		PrometheusConfig: application.Spec.Prometheus,
-	})
+	netpolSpec := networking.CreateNetPolSpec(
+		networking.NetPolOpts{
+			AccessPolicy:     application.Spec.AccessPolicy,
+			Ingresses:        &application.Spec.Ingresses,
+			Port:             &application.Spec.Port,
+			Namespace:        application.Namespace,
+			Name:             application.Name,
+			RelatedServices:  &egressServices,
+			PrometheusConfig: application.Spec.Prometheus,
+		},
+	)
 
 	if netpolSpec == nil {
 		r.SetControllerFinishedOutcome(ctx, application, controllerName, err)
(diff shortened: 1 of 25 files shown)


anderssonw:
> Unrelated to the changes in this PR, but is the istioEnabled flag on prometheus needed when the reconcile function can check the label on the namespace, which avoids making the users set the flag?

Hmm, thats a fair point. I'm not sure if @evenh had some reason for adding it to the prometheusConfig originally. Seemingly checking the namespace should be enough.

controllers/application/network_policy.go
```diff
@@ -2,61 +2,53 @@ package applicationcontroller
...
+			Namespace:        application.Namespace,
+			Name:             application.Name,
+			RelatedServices:  &egressServices,
+			PrometheusConfig: application.Spec.Prometheus,
+		},
+	)
 
-	networkPolicy := networkingv1.NetworkPolicy{ObjectMeta: metav1.ObjectMeta{Namespace: application.Namespace, Name: application.Name}}
-	_, err := ctrlutil.CreateOrPatch(ctx, r.GetClient(), &networkPolicy, func() error {
+	if netpolSpec == nil {
```
esphen, line 45: Have you taken into account the case where a user goes from having a networkpolicy to not having one? In other words, should there be a delete of orphaned netpols here?

samples/application.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   namespace: sample
   name: sample-one
 spec:
-  image: "image"
+  image: "quay.io/brancz/prometheus-example-app:v0.4.0"
```
esphen, line 7: Intentional change?

anderssonw: When testing locally, the "image" tag will cause some verbose errors which are somewhat of a nuisance. I could've chosen a "simpler" image, but we already know that this one works so me and Even used that for testing.

controllers/application/network_policy.go
anderssonw: Not sure I follow?

In the case that someone has a NetPol, but changes the Application in a way which makes the NetPol Spec empty, the original NetPol will be deleted. If the NetPol spec is not empty, it will be created by the `CreateOrPatch` function below

esphen: The first case you mentioned is what I'm referring to. Where the user used to have a netpol, but has changed the config so that it no longer should have one.

I see it now, it's nested as an argument to the function call below, so this seems fine. It might cause a few unecessary deletes though, GETs are generally cheaper as they can be cached. But that probably won't matter in the big picture.

```
r.GetClient().Delete(ctx, &networkPolicy)
```

esphen (approved):

controllers/application/network_policy.go
anderssonw: If you like I can put it in two steps, so it's easier to read what's going on :)

PR #273, opened 2023-07-25
Make inbound policy a pointer to allow for nil value

anderssonw:
The optional flag was removed for inbound rules in #272 (makes no sense to set inboundPolicy if no rules are added). It seems that due to the inbound policy not being a pointer, the required rule on the rules made the admission controller not allow applications. This was even the case when inboundPolicy was omitted. This seems to fix the issue as experienced by @BerntJulian


api/v1alpha1/podtypes/access_policy.go
```diff
@@ -3,13 +3,14 @@ package podtypes
 // +kubebuilder:object:generate=true
 type AccessPolicy struct {
 	//+kubebuilder:validation:Optional
-	Inbound InboundPolicy `json:"inbound,omitempty"`
+	Inbound *InboundPolicy `json:"inbound,omitempty"`
 	//+kubebuilder:validation:Optional
 	Outbound OutboundPolicy `json:"outbound,omitempty"`
```
esphen, line 8: Is there a similar issue with OutboundPolicy?

anderssonw: No rules on OutboundPolicy are required, so not as far as I can tell. @esphen

api/v1alpha1/podtypes/access_policy.go
```diff
@@ -3,13 +3,14 @@ package podtypes
...
 }
 
 // +kubebuilder:object:generate=true
 type InboundPolicy struct {
+	//+kubebuilder:validation:Required
 	Rules []InternalRule `json:"rules"`
```
esphen, line 14: Is adding required here a breaking change?

anderssonw: My presumption during the last PR was that no one has InboundPolicy set without also setting InternalRules, which would not make it breaking. However, the non-pointer on the InboundPolicy made it so it is somehow set in spite of not defining it in the Application Spec.

Currently it is already required, setting this flag just made it explicit, as the default behaviour in the API is required values unless otherwise specified

esphen (approved):

PR #275, opened 2023-07-26
[SKIP-705] More prepping for SKIPJobs

anderssonw:
* Moves test to designated folders to make it more readable when multiple different controllers are used
* More extrusion of functions to resourcegenerators


api/v1alpha1/application_types.go
```diff
@@ -127,19 +127,6 @@ type AuthorizationSettings struct {
 	AllowList []string `json:"allowList,omitempty"`
 }
 
-// ResourceRequirements
-// +kubebuilder:object:generate=true
-type ResourceRequirements struct {
```
esphen, line 132: I assume this was no longer in use?

controllers/application/egress_service_entry.go
```diff
@@ -22,138 +13,35 @@ func (r *ApplicationReconciler) reconcileEgressServiceEntry(ctx context.Context,
...
-	if err != nil {
-		r.SetControllerError(ctx, application, controllerName, err)
-		return reconcile.Result{}, err
-	}
+			r.SetLabelsFromApplication(ctx, &serviceEntry, *application)
+			util.SetCommonAnnotations(&serviceEntry)
 
-	for _, serviceEntry := range serviceEntries.Items {
-		// Skip unrelated service entries
-		if !isEgressServiceEntry(serviceEntry) {
```
esphen, line 71: I see this check is removed, I assume this is also because it is only iterating over owned resources?

controllers/application/egress_service_entry.go
```diff
@@ -22,138 +13,35 @@ func (r *ApplicationReconciler) reconcileEgressServiceEntry(ctx context.Context,
...
-	}
-
-	for _, port := range externalPorts {
-		if ruleIP == "" && port.Protocol == "TCP" {
-			r.GetRecorder().Eventf(
-				&application,
-				corev1.EventTypeWarning, "Invalid",
-				"A static IP must be set for TCP port %d",
-				port.Port,
-			)
```
esphen, line 131: It appears we removed this event from the service_entry.go. Are you confident that we'll be able to debug an issue related to this without it?

tests/application/access-policy/00-assert.yaml
```diff
@@ -59,7 +59,7 @@ spec:
 apiVersion: networking.istio.io/v1beta1
 kind: ServiceEntry
 metadata:
-  name: access-policy-egress-3a90cb5d70dc06a
+  name: access-policy-egress-3a90cb5d70dc06a-99561aeccea8b051
```
esphen, line 62: Will ServiceEntries change name? Could that somehow cause issues with orphaned serviceentries from the last version not being deleted by skiperator?

api/v1alpha1/application_types.go
anderssonw: It was not removed in the last PR, but should've been. It's also defined in the podtypes folder

controllers/application/egress_service_entry.go
anderssonw: Aye!

anderssonw: Or actually, no. In the deletion phase we list all the service entries in the namespace. However the check after, as defined in the `DeleteUnusedEgresses` function in the ReconcilerBase, checks for ownership either way. Result should be the same

tests/application/access-policy/00-assert.yaml
anderssonw: Yeah, that's definitely something to keep in mind. We should go over all the old service entries and delete those.

This is done due to the fact that jobs and applications would have the same naming scheme for access policies. One solution could be to append `job` or something if the type is job.

I'm not sure if you agree with how its done here, simply appending a hash of the name of the owned object.

esphen: ServiceEntries used to be prefixed with the application name (`skjema-bruksnavn-frontend-egress-5d0c98885286d5e1` for example), which I think would solve that issue? What's the reason we moved away from that? I feel a hash isn't as clear as user friendly as a plain text name

esphen: If you add the kind to that name I mean. Or what do you think, do you prefer a hash?

esphen: To prevent changing existing names you could even consider the application resources a "default", only adding a `job` prefix for jobs.

anderssonw: They are still prefixed with the application name. However, if a Job and an Application had the same name, there would be a naming collision for the service entry.

I think doing as you say with only pre/suffixing jobs is a solution, it just also feels sort of "unclean".. But it might be a better solution than having to clean up after *a lot* of old serviceentries

pkg/resourcegenerator/istio/service_entry.go
```diff
@@ -1,25 +1,42 @@
...
+			serviceEntryName := fmt.Sprintf("%s-egress-%x", object.GetName(), util.GenerateHashFromName(rule.Host))
+
+			objectKind := object.GetObjectKind().GroupVersionKind().Kind
+
+			switch object.(type) {
+			case *skiperatorv1alpha1.Application:
+				break
+			default:
+				serviceEntryName = fmt.Sprintf("%v-%v", strings.ToLower(objectKind), serviceEntryName)
+			}
```
anderssonw, line 31: SO to @evenboee for this one 😎

esphen (approved):

PR #291, opened 2023-08-17
Fix wordy "could not init status" error message on Application init

anderssonw:
Some weird interactions with finalizer currently, not sure why.

Would love to be able to remove the annoying 
```
{"controller": "skipjob", "controllerGroup": "skiperator.kartverket.no", "controllerKind": "SKIPJob", "SKIPJob": {"name":"condition-finish","namespace":"test"}, "namespace": "test", "name": "condition-finish", "reconcileID": "49b70cf2-d983-46be-b125-c8ae25b449e9", "error": "Operation cannot be fulfilled on skipjobs.skiperator.kartverket.no \"condition-finish\": StorageError: invalid object, Code: 4, Key: /registry/skiperator.kartverket.no/skipjobs/test/condition-finish, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: 23dff9d9-88ab-45e9-8a2f-798562f4960b, UID in object meta: "}
```

error was well, but not sure why it's happening. something going on when deleting applications

### DIFF
--- a/controllers/application/controller.go
+++ b/controllers/application/controller.go
@@ -54,7 +54,7 @@ const applicationFinalizer = "skip.statkart.no/finalizer"
 
 func (r *ApplicationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&skiperatorv1alpha1.Application{}).
+		For(&skiperatorv1alpha1.Application{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&appsv1.Deployment{}).
 		Owns(&corev1.Service{}).
 		Owns(&corev1.ConfigMap{}).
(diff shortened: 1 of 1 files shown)


controllers/application/controller.go
```diff
@@ -109,9 +107,59 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req reconcile.Req
...
+
+	if len(statusDiff) > 0 {
+		err := r.GetClient().Status().Update(ctx, application)
+		return reconcile.Result{Requeue: true}, err
+	}
+
+	if len(fullDiff) > 0 && len(statusDiff) == 0 {
+		err := r.GetClient().Update(ctx, application)
+		return reconcile.Result{Requeue: true}, err
+	}
```
anderssonw, line 156: this should not be necessary, but for some reason L#143-L#146 does not apply the finalizer...

anderssonw:
I have found that the 2nd annoying error log mentioned in the description happens after the initial `.Get()` function in `application/controller.go`. However, it is automatically logged upon Getting, which is somewhat annoying..

anderssonw:
According to some folks at the Kubernetes slack there doesn't seem to be any good solution for the error in the description as of now https://kubernetes.slack.com/archives/C0EG7JC6T/p1690885393706239

evenh (approved):

PR #292, opened 2023-08-18
Fix bug in mutable and shared annotations

evenh:
When a reconciliation loop adds a new annotation to the map it is silently reused for other apps reconciling at a later time.

This leads to a case where an app who doesn't have a Prometheus block specified gets attempted scraped at the previous reconciled app's (with Prometheus config) port and path.

FYI @christianmosveen 

SKIP-1134

### DIFF
--- a/controllers/application/deployment.go
+++ b/controllers/application/deployment.go
@@ -28,11 +28,7 @@ const (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


anderssonw (approved):
image

Silly little bug

anderssonw:
:shipit: ? @evenh

PR #295, opened 2023-08-23
SKIP-1084 Allow namespace-labels in AccessPolicies

SnorreSelmer:
Added test for using namespace-labels to generate AccessPolicy rules instead of managing large lists of namespaces manually

### DIFF
--- a/api/v1alpha1/podtypes/access_policy.go
+++ b/api/v1alpha1/podtypes/access_policy.go
@@ -1,5 +1,9 @@
 package podtypes
 
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
 // +kubebuilder:object:generate=true
 type AccessPolicy struct {
 	//+kubebuilder:validation:Optional
@@ -27,6 +31,8 @@ type InternalRule struct {
 	Namespace string `json:"namespace,omitempty"`
 	//+kubebuilder:validation:Required
 	Application string `json:"application"`
+	//+kubebuilder:validation:Optional
+	NamespacesByLabel *metav1.LabelSelector `json:"namespaceByLabel,omitempty"`
 }
 
 // +kubebuilder:object:generate=true
--- a/pkg/resourcegenerator/networking/network_policy.go
+++ b/pkg/resourcegenerator/networking/network_policy.go
@@ -185,9 +185,7 @@ func getInboundPolicyPeers(inboundRules []podtypes.InternalRule, namespace strin
 		}
 
 		policyPeers = append(policyPeers, networkingv1.NetworkPolicyPeer{
-			NamespaceSelector: &metav1.LabelSelector{
-				MatchLabels: map[string]string{"kubernetes.io/metadata.name": inboundRule.Namespace},
-			},
+			NamespaceSelector: getNamespaceSelector(inboundRule),
 			PodSelector: &metav1.LabelSelector{
 				MatchLabels: map[string]string{"app": inboundRule.Application},
 			},
@@ -195,6 +193,16 @@ func getInboundPolicyPeers(inboundRules []podtypes.InternalRule, namespace strin
 	}
 
 	return policyPeers
+}
+
+func getNamespaceSelector(rule podtypes.InternalRule) *metav1.LabelSelector {
+	if rule.NamespacesByLabel != nil {
+		return rule.NamespacesByLabel
+	} else {
+		return &metav1.LabelSelector{
+			MatchLabels: map[string]string{"kubernetes.io/metadata.name": rule.Namespace},
+		}
+	}
 }
 
 func hasExternalIngress(ingresses []string) bool {


SnorreSelmer:
Var litt kjapp på avtrekkeren med denne PR-en.. Må oppdatere app-definisjonen til å ta med det nye feltet, og gjøre ferdig tester.

api/v1alpha1/podtypes/access_policy.go
```diff
@@ -27,6 +31,8 @@ type InternalRule struct {
 	Namespace string `json:"namespace,omitempty"`
 	//+kubebuilder:validation:Required
 	Application string `json:"application"`
+	//+kubebuilder:validation:Optional
+	NamespacesByLabel *metav1.LabelSelector `json:"namespaceByLabel,omitempty"`
```
anderssonw, line 35: mismatch in the struct definition name and the json key

anderssonw (changes requested):
we have to keep the `Namespace` prop in the policy struct for backwards compatibility, so I think it would be easier for product teams to just provide a list of `AdditionalNamespaces` or something similar, which is just a list of strings, instead of a full on labelselector.

doing it like this would make it so teams have to write 
```
namespaceByLabel:
  kubernetes.io/metadata.name: somenamespace
```

according to https://kubernetes.io/docs/concepts/services-networking/network-policies/ it seems you can provide a list of namespaceselectors, so a function could build the list according to the contents of accessPolicy.namespace and accesspolicy.additionalNamepsaces or similar

also make sure you add tests in `tests/application/access-policy` which check for this change!

anderssonw:
> Var litt kjapp på avtrekkeren med denne PR-en.. Må oppdatere app-definisjonen til å ta med det nye feltet, og gjøre ferdig tester.

@SnorreSelmer make sure you always run `make build` locally before pushing! this ensures that the CRD and `zz_generated_deepcopy.go` is generated, and that Skiperator is able to be built :)

SnorreSelmer:
> > Var litt kjapp på avtrekkeren med denne PR-en.. Må oppdatere app-definisjonen til å ta med det nye feltet, og gjøre ferdig tester.
> 
> @SnorreSelmer make sure you always run `make build` locally before pushing! this ensures that the CRD and `zz_generated_deepcopy.go` is generated, and that Skiperator is able to be built :)

Yep. @esphen and I are going through building tests etc. now.

api/v1alpha1/podtypes/access_policy.go
SnorreSelmer: Thanks, fixed

esphen:
It wasn't clear to me what the issue with backwards compatibility was, so I tried to write something down. Consider the following cases:

- Existing scenario: User has namespace defined in policy, controller uses defined namespace
- Existing scenario: User has no namespace defined in policy, controller uses current namespace
- New scenario: User has namespaceByLabels defined in policy, controller uses the label selector defined
- New scenario: Use has both namespaceByLabel and namespace defined in policy. This needs to be discussed, but for example it could thrown an error

Only the top two are present in the cluster, the two new scenarios should not affect the existing Application manifests.

Did I get that right? Am I missing something?

Also, a full on labelselector is more flexible as it allows using existing labels like `team` and new labels we (or potentially ones the teams) create in the future, not just namespace names.

anderssonw (commented):
As it is now I think it looks fine. There should be no issues with backwards compatibility, I only mentioned for before me and @SnorreSelmer talked about what this was *actually* fixing. I believe this looks fine, but I agree that we need to talk about how we should act if both namespace and namespacebylabel is set.

IMO, setting both should use `Namespace`, as that should give the least amount of privilege to a poorly set configuration, which sounds safer in my book. However, I don't think throwing an error is necessary, perhaps only sending an event? The product team using it still has to know that it happened, and sometimes that's not super easy

api/v1alpha1/podtypes/access_policy.go
```diff
@@ -1,5 +1,9 @@
 package podtypes
 
+//import (
+//	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+//)
+
```
anderssonw, line 6: Please remove unused comments

api/v1alpha1/podtypes/access_policy.go
```diff
@@ -22,11 +26,15 @@ type OutboundPolicy struct {
 	External []ExternalRule `json:"external,omitempty"`
 }
 
+// If user adds both Namespace and NamespacesByLabel to an InternalRule, Namespace takes presedence
+// +kubebuilder:object:generate=true
```
anderssonw, line 30: this is _super_ nitpicky, but this sort of sounds like internal documentation and not documentation for the API

maybe some more natural wording would be along the lines of "Adding both Namespace and NamespacesByLabel to InternalRule will cause Namespace to take precedence, and NamespacesByLabel will be omitted"

pkg/resourcegenerator/networking/network_policy.go
```diff
@@ -197,6 +195,22 @@ func getInboundPolicyPeers(inboundRules []podtypes.InternalRule, namespace strin
...
+			MatchLabels: map[string]string{"kubernetes.io/metadata.name": rule.Namespace},
+		}
+	} else if rule.NamespacesByLabel != nil {
+		return &metav1.LabelSelector{
+			MatchLabels: rule.NamespacesByLabel,
+		}
+	} else {
+		return &metav1.LabelSelector{
+			MatchLabels: map[string]string{"kubernetes.io/metadata.name": rule.Namespace},
+		}
```
anderssonw, line 210: this should never occur due to L#183-L#185

perhaps it would make more sense to send the application namespace to `getNamespaceSelector()` and then have the catchAll be

```go
return &metav1.LabelSelector{
	MatchLabels: map[string]string{"kubernetes.io/metadata.name": applicationNamespace},
}
```

tests/application/namespace-label/00-application.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: skiperator.kartverket.no/v1alpha1
+kind: Application
+metadata:
+  name: test-source
```
anderssonw, line 4: nice if the application has a descriptive name like `namespace-label`

tests/application/namespace-label/00-application.yaml
```diff
@@ -0,0 +1,14 @@
...
+spec:
+  image: image
+  port: 8080
+  accessPolicy:
+    inbound:
+      rules:
+        - application: test-ext
+          namespacesByLabel:
+            team: skiptest
+            app: testapp-selsno
```
anderssonw, line 14: `selsno`?

anderssonw (changes requested):
lotta nitpicks, everything looks pretty good though :)

tests/application/namespace-label/00-application.yaml
SnorreSelmer: **Sel**mer, **Sno**rre < My name. Just for some uniqueness

anderssonw: IMO don't use your own name in the tests @SnorreSelmer, find something else to keep it unique

tests/application/namespace-label/00-application.yaml
SnorreSelmer: Building a new iteration with more descriptive app-names

api/v1alpha1/podtypes/access_policy.go
```diff
@@ -22,11 +22,16 @@ type OutboundPolicy struct {
...
 type InternalRule struct {
-	//+kubebuilder:validation:Optional
-	Namespace string `json:"namespace,omitempty"`
 	//+kubebuilder:validation:Required
 	Application string `json:"application"`
+	//+kubebuilder:validation:Optional
+	Namespace string `json:"namespace,omitempty"`
+	//+kubebuilder:validation:Optional
+	NamespacesByLabel map[string]string `json:"namespacesByLabel,omitempty"`
```
anderssonw, line 34: nice! one last thing is maybe documenting what Namespace does and what NamespacesByLabel does, also adding this to the README

worth adding to readme seeing as we still don't have skiperator released to pkg.go.dev quite yet :)

anderssonw (approved):
final comment, approved as its just docs remaining

api/v1alpha1/podtypes/access_policy.go
SnorreSelmer: Sorted in the commit below

anderssonw:
Hate to be that guy @SnorreSelmer but with the new changes from #226 you might as well remove from README and update docs.. Sorry 😅

SnorreSelmer:
That’s ok

fre. 29. sep. 2023 kl. 11:54 skrev William Andersson <
***@***.***>:

esphen (approved):

PR #296, opened 2023-08-24
SKIP-1147 Fix patching of ServiceEntries when updating external accessPolicy

omaen:
Discovered that the ServiceEntry resources wasn't updated when changes were made to the accessPolicy in the skiperator manifest. This is due to how the `CreateOrPatch` function excpects changes to be made on the object inside the mutating function. 

Although a bit obscure, the docstring for `CreateOrPatch` states:
> The object's desired state must be reconciled with the before state inside the passed in callback MutateFn

We are reconstructing the serviceEntries before the CreateOrPatch function, therefore we need to send in the correct copy.


anderssonw:
BTW @omaen, perhaps it would be possible to create a kuttl test which modifies an Applications external rules so we have tests to avoid something like this again?

evenh (approved):

PR #308, opened 2023-09-05
Added labels to deployment pod template spec

BardOve:
Labels field in application definition now propegate correctly to pod template in deployment.

Changes GroupKindFromControllerResource to a map instead of a function as it was essentially already functioning as one.

Uses the standard SetResourceLabelIfApplies by adding Pod to the ControllerResources and adding a map from the resource -> GroupKind.

Had to create a pod and extract the template, as PodTemplateSpec does not implement type Object, which is needed for setting labels.

### DIFF
--- a/api/v1alpha1/application_types.go
+++ b/api/v1alpha1/application_types.go
@@ -9,7 +9,6 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	"strings"
 	"time"
 )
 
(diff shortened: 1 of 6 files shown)


esphen (approved):
Funksjonelt sett ser det greit ut, men hadde vært fint om vi fikk de-duplisert de nye labels-funksjonene etter hvert

anderssonw (commented):
Does PodTemplateSpec not implement `.GetObjectKind().GroupVersionKind()`? Is there a way we can instead build a Pod object and use its Spec if not?

This seems like a circumvention of refactoring the Deployment resource generating, to avoid using already existing function

anderssonw:
Added a fix according to what I specified. I'm not a huge fan of how this thing works, and would honestly prefer if i could somehow get all owned resources and apply it in the main controller function instead of separately for each resource. Asked in the controller-runtime slack if something like that is feasible, so we will see :)

anderssonw (approved):
lgtm

PR #321, opened 2023-09-21
Add support for ignoring reconciliation of children

esphen:
In some cases it may be necessary to manually edit child resources of an application. This is usually in the short term in terms of debugging a problem or when implementing a temporary workaround. In this case we need the latter and instead of migrating the team out of their skiperator app we add a feature here to allow us to edit the Gateway resource temporarily.

This PR adds a feature where you can manually label a child with `skiperator.kartverket.no/ignore: "true"` to opt that resource out of reconciliation. When you are done with that resource, you simply remove the label. Note that this is not something you can do via `resourceLabels`, as skiperator will add the label but because of the resource then being opted out of reconcile it will not remove the label.

I forsee this feature as a SKIP-team feature where we will mostly use it ourselves, perhaps even blocking the teams from setting this label.

### DIFF
--- a/controllers/application/controller.go
+++ b/controllers/application/controller.go
@@ -3,9 +3,10 @@ package applicationcontroller
 import (
 	"context"
 	"fmt"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"regexp"
 	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	policyv1 "k8s.io/api/policy/v1"
 
@@ -53,22 +54,77 @@ const applicationFinalizer = "skip.statkart.no/finalizer"
 
 func (r *ApplicationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&skiperatorv1alpha1.Application{}).
-		Owns(&appsv1.Deployment{}).
-		Owns(&corev1.Service{}).
-		Owns(&corev1.ConfigMap{}).
-		Owns(&networkingv1beta1.ServiceEntry{}).
+		For(&skiperatorv1alpha1.Application{}, builder.WithPredicates(
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
+		)).
+		Owns(&appsv1.Deployment{}, builder.WithPredicates(
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
+		)).
+		Owns(&corev1.Service{}, builder.WithPredicates(
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
+		)).
+		Owns(&corev1.ConfigMap{}, builder.WithPredicates(
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
+		)).
+		Owns(&networkingv1beta1.ServiceEntry{}, builder.WithPredicates(
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
+		)).
 		Owns(&networkingv1beta1.Gateway{}, builder.WithPredicates(
 			util.MatchesPredicate[*networkingv1beta1.Gateway](isIngressGateway),
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
 		)).
-		Owns(&autoscalingv2.HorizontalPodAutoscaler{}).
-		Owns(&networkingv1beta1.VirtualService{}).
-		Owns(&securityv1beta1.PeerAuthentication{}).
-		Owns(&corev1.ServiceAccount{}).
-		Owns(&policyv1.PodDisruptionBudget{}).
-		Owns(&networkingv1.NetworkPolicy{}).
-		Owns(&securityv1beta1.AuthorizationPolicy{}).
-		Owns(&pov1.ServiceMonitor{}).
+		Owns(&autoscalingv2.HorizontalPodAutoscaler{}, builder.WithPredicates(
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
+		)).
+		Owns(&networkingv1beta1.VirtualService{}, builder.WithPredicates(
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
+		)).
+		Owns(&securityv1beta1.PeerAuthentication{}, builder.WithPredicates(
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
+		)).
+		Owns(&corev1.ServiceAccount{}, builder.WithPredicates(
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
+		)).
+		Owns(&policyv1.PodDisruptionBudget{}, builder.WithPredicates(
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
+		)).
+		Owns(&networkingv1.NetworkPolicy{}, builder.WithPredicates(
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
+		)).
+		Owns(&securityv1beta1.AuthorizationPolicy{}, builder.WithPredicates(
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
+		)).
+		Owns(&pov1.ServiceMonitor{}, builder.WithPredicates(
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
+		)).
 		Watches(&certmanagerv1.Certificate{}, handler.EnqueueRequestsFromMapFunc(r.SkiperatorOwnedCertRequests)).
 		WithEventFilter(predicate.GenerationChangedPredicate{}).
 		Complete(r)
--- a/controllers/application/controller.go
+++ b/controllers/application/controller.go
@@ -54,77 +54,22 @@ const applicationFinalizer = "skip.statkart.no/finalizer"
 
 func (r *ApplicationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&skiperatorv1alpha1.Application{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&appsv1.Deployment{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&corev1.Service{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&corev1.ConfigMap{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&networkingv1beta1.ServiceEntry{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
+		For(&skiperatorv1alpha1.Application{}).
+		Owns(&appsv1.Deployment{}).
+		Owns(&corev1.Service{}).
+		Owns(&corev1.ConfigMap{}).
+		Owns(&networkingv1beta1.ServiceEntry{}).
 		Owns(&networkingv1beta1.Gateway{}, builder.WithPredicates(
 			util.MatchesPredicate[*networkingv1beta1.Gateway](isIngressGateway),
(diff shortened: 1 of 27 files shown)


controllers/application/controller.go
```diff
@@ -53,22 +54,77 @@ const applicationFinalizer = "skip.statkart.no/finalizer"
...
-		For(&skiperatorv1alpha1.Application{}).
-		Owns(&appsv1.Deployment{}).
-		Owns(&corev1.Service{}).
-		Owns(&corev1.ConfigMap{}).
-		Owns(&networkingv1beta1.ServiceEntry{}).
+		For(&skiperatorv1alpha1.Application{}, builder.WithPredicates(
+			predicate.Not(
+				util.MatchesPredicate[client.Object](hasIgnoreLabel),
+			),
+		)).
```
esphen, line 61: Please give input on how we can remove duplication here 🙂

evenh: A somewhat cleaner approach may be:

```patch
Subject: [PATCH] Foo
---
Index: pkg/util/predicates.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/pkg/util/predicates.go b/pkg/util/predicates.go
--- a/pkg/util/predicates.go	(revision 406d3d39a0fc6e624cb71bfc579a0f93816b9560)
+++ b/pkg/util/predicates.go	(date 1695325207682)
@@ -1,8 +1,16 @@
 package util
 
 import (
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+var ShouldBeIgnored = builder.WithPredicates(
+	predicate.Not(MatchesPredicate[client.Object](func(object client.Object) bool {
+		return object.GetLabels()[SkiperatorIgnoreLabel] == "true"
+	})),
 )
 
 type MatchesPredicate[T client.Object] func(T) bool
Index: pkg/util/constants.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/pkg/util/constants.go b/pkg/util/constants.go
--- a/pkg/util/constants.go	(revision 406d3d39a0fc6e624cb71bfc579a0f93816b9560)
+++ b/pkg/util/constants.go	(date 1695325091458)
@@ -20,6 +20,8 @@
 	IstioMetricsPath       = "/stats/prometheus"
 
 	IstioRevisionLabel = "istio.io/rev"
+
+	SkiperatorIgnoreLabel = "skiperator.kartverket.no/ignore"
 )
 
 // A security context for use in pod containers created by Skiperator
Index: controllers/application/ingress_gateway.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/controllers/application/ingress_gateway.go b/controllers/application/ingress_gateway.go
--- a/controllers/application/ingress_gateway.go	(revision 406d3d39a0fc6e624cb71bfc579a0f93816b9560)
+++ b/controllers/application/ingress_gateway.go	(date 1695325212458)
@@ -134,8 +134,3 @@
 
 	return match
 }
-
-func hasIgnoreLabel(obj client.Object) bool {
-	labels := obj.GetLabels()
-	return labels["skiperator.kartverket.no/ignore"] == "true"
-}
Index: controllers/application/controller.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/controllers/application/controller.go b/controllers/application/controller.go
--- a/controllers/application/controller.go	(revision 406d3d39a0fc6e624cb71bfc579a0f93816b9560)
+++ b/controllers/application/controller.go	(date 1695325372140)
@@ -3,18 +3,6 @@
 import (
 	"context"
 	"fmt"
-	"regexp"
-	"strings"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	policyv1 "k8s.io/api/policy/v1"
-
-	certmanagerv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
-	skiperatorv1alpha1 "github.com/kartverket/skiperator/api/v1alpha1"
-	"golang.org/x/exp/maps"
-
-	"github.com/kartverket/skiperator/pkg/util"
 	pov1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	networkingv1beta1 "istio.io/client-go/pkg/apis/networking/v1beta1"
 	securityv1beta1 "istio.io/client-go/pkg/apis/security/v1beta1"
@@ -22,10 +10,20 @@
 	autoscalingv2 "k8s.io/api/autoscaling/v2"
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
+	policyv1 "k8s.io/api/policy/v1"
+	"regexp"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	certmanagerv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
+	skiperatorv1alpha1 "github.com/kartverket/skiperator/api/v1alpha1"
+	"golang.org/x/exp/maps"
+
+	"github.com/kartverket/skiperator/pkg/util"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -54,77 +52,20 @@
 
 func (r *ApplicationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&skiperatorv1alpha1.Application{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&appsv1.Deployment{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&corev1.Service{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&corev1.ConfigMap{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&networkingv1beta1.ServiceEntry{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&networkingv1beta1.Gateway{}, builder.WithPredicates(
-			util.MatchesPredicate[*networkingv1beta1.Gateway](isIngressGateway),
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&autoscalingv2.HorizontalPodAutoscaler{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&networkingv1beta1.VirtualService{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&securityv1beta1.PeerAuthentication{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&corev1.ServiceAccount{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&policyv1.PodDisruptionBudget{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&networkingv1.NetworkPolicy{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&securityv1beta1.AuthorizationPolicy{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
-		Owns(&pov1.ServiceMonitor{}, builder.WithPredicates(
-			predicate.Not(
-				util.MatchesPredicate[client.Object](hasIgnoreLabel),
-			),
-		)).
+		For(&skiperatorv1alpha1.Application{}, util.ShouldBeIgnored).
+		Owns(&appsv1.Deployment{}, util.ShouldBeIgnored).
+		Owns(&corev1.Service{}, util.ShouldBeIgnored).
+		Owns(&corev1.ConfigMap{}, util.ShouldBeIgnored).
+		Owns(&networkingv1beta1.ServiceEntry{}, util.ShouldBeIgnored).
+		Owns(&networkingv1beta1.Gateway{}, util.ShouldBeIgnored).
+		Owns(&autoscalingv2.HorizontalPodAutoscaler{}, util.ShouldBeIgnored).
+		Owns(&networkingv1beta1.VirtualService{}, util.ShouldBeIgnored).
+		Owns(&securityv1beta1.PeerAuthentication{}, util.ShouldBeIgnored).
+		Owns(&corev1.ServiceAccount{}, util.ShouldBeIgnored).
+		Owns(&policyv1.PodDisruptionBudget{}, util.ShouldBeIgnored).
+		Owns(&networkingv1.NetworkPolicy{}, util.ShouldBeIgnored).
+		Owns(&securityv1beta1.AuthorizationPolicy{}, util.ShouldBeIgnored).
+		Owns(&pov1.ServiceMonitor{}, util.ShouldBeIgnored).
 		Watches(&certmanagerv1.Certificate{}, handler.EnqueueRequestsFromMapFunc(r.SkiperatorOwnedCertRequests)).
 		WithEventFilter(predicate.GenerationChangedPredicate{}).
 		Complete(r)
```

Or even applying `util.ShouldBeIgnored` to just the `For` and `WithEventFilter` blocks. Side quest: should we handle cert manager resources?

evenh: Either way it needs some tests in order to verify functionality

esphen: Not sure if we can use the kuttl tests to test this as it requires multiple steps to verify, not just apply-check?

Ideally:

1. Apply application
2. Wait for resources to be created
3. Verify changing resource does revert change
4. Label subresource with ignore
5. Verify changing subresource does not revert change
6. Remove ignore label
7. Verify changing resource does revert change

evenh: Feel free to chime in if it's feasible @anderssonw

anderssonw: That sounds possible to test with Kuttl from what I can tell. The ii-assert.yaml files don't just work as "the cluster should contain this", but also "wait for the cluster to contain this". Its just gonna be a pretty long test.

Though we have had issues with testing mutating resources with Kuttl in the past, I think it's worth a shot from what I can tell

anderssonw: Also adding all of these event filters is essentially the same as checking the request object in the top level `controller.go` file, and checking if the object that sent the reconcile request has the label. The only difference is that the reconcile would start first instead of being ignored. It should be _a lot_ less wordy to skip the reconciles, though. There might be some way of creating your own eventfilters as @evenh mentioned, but I'm not familiar with it

controllers/application/ingress_gateway.go
```diff
@@ -134,3 +134,8 @@ func isIngressGateway(gateway *networkingv1beta1.Gateway) bool {
 
 	return match
 }
+
+func hasIgnoreLabel(obj client.Object) bool {
+	labels := obj.GetLabels()
+	return labels["skiperator.kartverket.no/ignore"] == "true"
+}
```
anderssonw, line 141: This should probably be placed somewhere else instead of under `ingress_gateway.go`

evenh: TODO: Check naming convention

anderssonw:
@esphen With the previous setup, only requests for reconciliation for the Application were ignored if the object with ignoreLabel was changed. If a different object was changed and requested a reconciliation, the reconcile would still happen on the object itself. The previous commit is one way of actually ignoring the reconcile request.

This _works_ but I find it a terribly poor solution. It's inconsistent between objects, it's wordy, and it's repeated for every single reconciled object. If we could somehow receive the object sending the request in the main `Reconcile` function in `controller.go`, this could be simplified hugely. However, I'm not entirely sure if we can..

anderssonw:
I want to look into this more, but if this was a request regarding se-eiendoms monday deployment its fair we want to just push out a working solution. I'm writing a few tests as we speak.

anderssonw:
> @esphen With the previous setup, only requests for reconciliation for the Application were ignored if the object with ignoreLabel was changed. If a different object was changed and requested a reconciliation, the reconcile would still happen on the object with the ignore label. The previous commit is one way of actually ignoring the reconcile request on the object with the label.
> 
> This _works_ but I find it a terribly poor solution. It's inconsistent between objects, it's wordy, and it's repeated for every single reconciled object. If we could somehow receive the object sending the request in the main `Reconcile` function in `controller.go`, this could be simplified hugely. However, I'm not entirely sure if we can..

Actually, never mind. Receiving the object that sent the Request won't be much help either, much due to the same same problem as the initial commit on this PR.

evenh (changes requested):
Looks good overall. However we need to fix the tests locally so we don't introduce new bugs

BardOve (approved):

PR #342, opened 2023-10-26
Set UnhealthyPodEviction to AlwaysAllow

omaen:
Enabled i kubernetes 1.27, but supported in 1.26 spec.

From the kubernetes docs:
> It is recommended to set AlwaysAllow [Unhealthy Pod Eviction Policy](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#unhealthy-pod-eviction-policy) to your PodDisruptionBudgets to support eviction of misbehaving applications during a node drain. The default behavior is to wait for the application pods to become [healthy](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#healthiness-of-a-pod) before the drain can proceed.


anderssonw (approved):
nice find :)

anderssonw:
This is ready for merge, @omaen? Deploying some version bumps so might as well grab everything else and have at it

anderssonw:
:shipit:

PR #358, opened 2023-11-20
[SKIP-1153] - refactor from kuttl to chainsaw

martinhny:
- Setting up a local environment can now be done with `make setup-local`. This was required to run tests in a workflow
- Tests can now run concurrently. We should try to make it so tests can run concurrently. This means no hardcoded namespaces unless strictly neccessary. Chainsaw creates unique namespaces for us if the metadata namespace field isnt provided. Hardcoded namespaces should have a unique prefix (directory name for example, ie podmonitor-ns)
- Delete steps aren't required anymore as chainsaw will delete all resources anyway
- Instead of the 00,01 syntax we now use brief but descriptive test file names. Every test directory should have a chainsaw-test.yaml file which decides the test steps.


eddycharly:
> Future tests should be made so they can run concurrently

If some tests need to run sequentially you can set `concurrent: false` in the `Test` spec.
Those tests will run first (sequentially) then other tests will run next (concurrently).

anderssonw:
:eyes:

BardOve (approved):

PR #359, opened 2023-11-21
Fix "resources" not getting updated

evenh:
Should be rolled out slowly (e.g. sandbox first) in order to verify that we're not introducing another reconciliation loop.

See SKIP-1245.


martinhny (approved):

evenh:
I think we might go down that route in the future, but for now let's keep the old approach as well and monitor how it works.

PR #360, opened 2023-11-22
SKIP-1225 Egress using NamespacesByLabel

SnorreSelmer:
accessPolicy now supports namespacesByLabel on outbound as well as inbound rules.


pkg/resourcegenerator/networking/network_policy.go
```diff
@@ -58,20 +59,24 @@ func getPolicyTypes(ingressRules []networkingv1.NetworkPolicyIngressRule, egress
...
 	}
 
 	for _, outboundRule := range (*accessPolicy).Outbound.Rules {
-		if outboundRule.Namespace == "" {
+		if outboundRule.Namespace == "" && outboundRule.NamespacesByLabel == nil {
 			outboundRule.Namespace = namespace
 		}
 
-		relatedService, isApplicationAvailable := getRelatedService(availableServices, outboundRule)
+		relatedService, isApplicationAvailable := getRelatedService(availableServices, outboundRule, namespaces)
```
martinhny, line 79: maybe stupid question, but why do we need this step? wont it return the same result as `GetEgressServices`, so the same values that already exist in `availableServices`?

esphen: Not sure. May be that can be optimised.

What I can see is that`GetEgressServices` takes an `accessPolicy`, whereas `getRelatedService` takes an `InternalRule`. Here we are working in the context of a single rule instead of the entire access policy. So it seems that this function further filters the result set into a finer subset. That may be significant.

tests/application/namespace-tests/00-application.yaml
esphen, line 1: Why is namespace-tests deleted?

pkg/util/reconciler.go
```diff
@@ -85,29 +86,68 @@ func (r *ReconcilerBase) GetEgressServices(ctx context.Context, owner client.Obj
 	}
 
 	for _, outboundRule := range accessPolicy.Outbound.Rules {
-		if outboundRule.Namespace == "" {
-			outboundRule.Namespace = owner.GetNamespace()
-		}
+		namespaces := []string{}
+
+		if outboundRule.NamespacesByLabel == nil {
```
esphen, line 91: On a general note - what happens when a user specifies both NamespacesByLabel and Namespace in a rule? Should we add some validation and give the user some feedback if that happens? And maybe add a test?

tests/application/namespace-tests/00-application.yaml
martinhny: afaik its moved to access-policies, which is more correct imo, as it just tests access-policy features.

SnorreSelmer: Correct.

pkg/util/reconciler.go
SnorreSelmer: Yeah, we should. Namespace has priority over NamespacesByLabel, as stated in the documentation. The code reflects this, but there is no feedback anywhere if both are supplied

pkg/resourcegenerator/networking/network_policy.go
```diff
@@ -106,15 +109,33 @@ func getEgressRules(accessPolicy *podtypes.AccessPolicy, namespace string, avail
...
+				for _, namespace := range namespaces.Items {
+					labels := namespace.GetLabels()
+					if labels != nil {
+						for key, value := range labels {
+							if rule.NamespacesByLabel[key] == value {
+								return service, true
+							}
+						}
+					}
+				}
```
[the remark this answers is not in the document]
SnorreSelmer: Fixed in https://github.com/kartverket/skiperator/commit/48eafcd579d27c8197a57fa4f64d81e48baed99a

controllers/skipjob/network_policy.go
```diff
@@ -17,9 +18,16 @@ func (r *SKIPJobReconciler) reconcileNetworkPolicy(ctx context.Context, skipJob
 		return reconcile.Result{}, err
 	}
 
+	namespaces, err := r.GetNamespaces(ctx, skipJob)
+	if err != nil {
+		//r.SetControllerFinishedOutcome(ctx, skipJob, controllerName, err)
```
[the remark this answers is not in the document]
SnorreSelmer: Fixed in https://github.com/kartverket/skiperator/commit/1c6e488f11b99c3bc89d566e1568d6472ada2444

pkg/util/reconciler.go
SnorreSelmer: Fixed in https://github.com/kartverket/skiperator/commit/42830ccf3e463a8e01fd291f9f91d1092e4b3521

esphen (approved):
:crossed_fingers:

berg-thom (approved):

PR #366, opened 2023-12-08
Bump various dependencies

evenh:


api/v1alpha1/skipjob_defaults.go
```diff
@@ -1,7 +1,7 @@
 package v1alpha1
 
 import (
-	"github.com/imdario/mergo"
+	"dario.cat/mergo"
```
evenh, line 4: Package relocation

berg-thom (approved):

PR #368, opened 2023-12-11
[SKIP-1185] - add missing securityContext values in deployment

martinhny:


esphen (approved):
There is a possibility that this causes a regression for some apps, so this should be tested with some example apps before wider deployment

evenh (approved):

PR #372, opened 2023-12-11
Add more permissions for proxy, so frontends can run

martinhny:
The Caddy dockerfile sets the net_bind_service capability, so even if we don't really require it for our usecase, the docker image requires it to run. Without it many frontends will not work.


esphen:
What's cool about this is that it's still compliant with restricted PSA, because the NET_BIND_SERVICE is the only capability that is allowed to be re-added! Probably for good reason :slightly_smiling_face: 

https://kubernetes.io/docs/concepts/security/pod-security-standards/

esphen (approved):

PR #373, opened 2023-12-12
Workflow should fail if tests fail

martinhny:


Makefile
```diff
@@ -112,10 +114,11 @@ run-test: build
 	PID=$$!; \
 	echo "skiperator PID: $$PID"; \
 	echo "Log redirected to file: $$LOG_FILE"; \
-	if [ -z "$(TEST_DIR)" ]; then \
-		$(MAKE) test; \
-	else \
-		$(MAKE) test-single dir=$(TEST_DIR); \
-	fi; \
-	echo "Stopping skiperator (PID $$PID)..."; \
-	kill $$PID
```
evenh, line 121: We still need to kill skiperator manually

evenh (changes requested):

Makefile
martinhny: Seems like its catching the wrong PID, at least for me. Ill try to fix it

martinhny: Added it back and tested fine now. Think something with me was just not working properly.

Makefile
```diff
@@ -112,10 +114,11 @@ run-test: build
...
+	( \
+		if [ -z "$(TEST_DIR)" ]; then \
+			$(MAKE) test; \
+		else \
+			$(MAKE) test-single dir=$(TEST_DIR); \
+		fi; \
+	) && \
+	(kill $$PID && echo "Stopping skiperator (PID $$PID)...")  || (echo "Test or skiperator failed" && exit 1)
```
evenh, line 124:
```shell
(echo "Stopping skiperator (PID $$PID)..." && kill $$PID)
```
So we're noticing which PID is about to be killed

martinhny: Changed the order, also added kill when the test fail

evenh (approved):

martinhny:
Just a note about this PR; This will cause our test workflow to occasionally fail due to concurrency issues in skiperator. Rerunning the workflow should fix it.

berg-thom (approved):

PR #384, opened 2023-12-20
application: add 'team' to spec

evenh:
In order to propagate team name down to running Pod's this field needs to be explicitly set or inferred from the Application's namespace.

### DIFF
--- a/controllers/application/controller.go
+++ b/controllers/application/controller.go
@@ -3,6 +3,7 @@ package applicationcontroller
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/types"
 	"regexp"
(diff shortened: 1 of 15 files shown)


controllers/application/controller.go
```diff
@@ -122,6 +123,13 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req reconcile.Req
 		application.Labels = aggregateLabels
 	}
 
+	// Add team label
+	if len(application.Spec.Team) == 0 {
```
martinhny, line 127: I think we should add a testcase for an application in a namespace without the team label.

martinhny (approved):

PR #390, opened 2024-01-09
Only scrape istio-proxy

evenh:
We'll always want to scrape istio-proxy. If custom application metrics is exposed through the spec, the istio-proxy should scrape those metrics and expose the merged result.


controllers/application/service_monitor.go
```diff
@@ -32,17 +31,6 @@ func (r *ApplicationReconciler) reconcileServiceMonitor(ctx context.Context, app
 		return util.RequeueWithError(err)
 	}
 
-	if application.Spec.Prometheus == nil {
```
esphen, line 35: What happens if you toggle istio on and off? Should this be changed to `!r.IsIstioEnabledForNamespace(ctx, application.Namespace)` instead of being removed so that it cleans up properly?

evenh: I was considering the possibility of disabling istio in a Skiperator-managed application namespace rather slim, but I could absolutely re-add this block in order to lighten the load on grafana-agent when there's nothing to scrape.

esphen (approved):

PR #395, opened 2024-01-16
Use correct labelselector for skipjob pods in netpol

anderssonw:
Uses the correct labelSelector for SKIPJobs, as the Pods are labeled with `app: nameOfSKIPJob-skipjob`.

Also adds a little line regarding accessPolicy names for SKIPJobs


evenh (approved):
Nice 👏🏻 The Docker push currently fails, but probably due the PR coming from a fork.

PR #405, opened 2024-02-07
Allow traffic to new internal server network

omaen:
Needed to allow outbound traffic to new databases.

### DIFF
--- a/controllers/namespace/default_deny_network_policy.go
+++ b/controllers/namespace/default_deny_network_policy.go
@@ -28,9 +28,15 @@ func (r *NamespaceReconciler) reconcileDefaultDenyNetworkPolicy(ctx context.Cont
 				networkingv1.PolicyTypeEgress,
 			},
 			Egress: []networkingv1.NetworkPolicyEgressRule{
-				// Egress rule for Internet
 				{
 					To: []networkingv1.NetworkPolicyPeer{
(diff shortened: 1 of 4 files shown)


evenh (approved):

controllers/namespace/default_deny_network_policy.go
```diff
@@ -28,9 +28,15 @@ func (r *NamespaceReconciler) reconcileDefaultDenyNetworkPolicy(ctx context.Cont
...
 			},
 			Egress: []networkingv1.NetworkPolicyEgressRule{
-				// Egress rule for Internet
 				{
 					To: []networkingv1.NetworkPolicyPeer{
+						// Egress rule for parts of internal server network
+						{
+							IPBlock: &networkingv1.IPBlock{
+								CIDR: "10.40.0.0/16",
+							},
```
esphen, line 37: Assuming this is the subnet where databases live, correct me if I'm wrong. Is this allowing traffic here by default? Should we make the product teams open this in their Skiperator manifests?

esphen (approved):

controllers/namespace/default_deny_network_policy.go
esphen: We took this offline and reviewed the security aspects of this change. With the current setup this should work fine

PR #415, opened 2024-02-23
Path based routing

omaen:
Extended skiperator API with new type: `Routing` which is responsible for setting up ingress for one or more applications. This new controller creates Certificate, Gateway, Network Policy and VirtualService resources to make this happen.

Should be compatible with normal Skiperator Applications, but the migration path is somewhat unexplored, because it would require the application to remove its ingresses at the same time the Routing resource is created.

### DIFF
--- a/api/v1alpha1/routing_types.go
+++ b/api/v1alpha1/routing_types.go
@@ -23,9 +23,7 @@ type Routing struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	//+kubebuilder:validation:Required
-	Spec RoutingSpec `json:"spec,omitempty"`
-
-	//+kubebuilder:validation:Optional
+	Spec   RoutingSpec   `json:"spec,omitempty"`
 	Status RoutingStatus `json:"status,omitempty"`
 }
 
@@ -78,4 +76,12 @@ func (in *Routing) GetVirtualServiceName() string {
 func (in *Routing) GetCertificateName() (string, error) {
 	namePrefix := fmt.Sprintf("%s-%s", in.Namespace, in.Name)
 	return util.GetSecretName(namePrefix, "routing-ingress")
+}
+
+func (in *Routing) GetConditions() []metav1.Condition {
+	return in.Status.Conditions
+}
+
+func (in *Routing) SetConditions(conditions []metav1.Condition) {
+	in.Status.Conditions = conditions
 }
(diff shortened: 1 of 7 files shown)


api/v1alpha1/routing_types.go
```diff
@@ -0,0 +1,64 @@
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+//+kubebuilder:object:root=true
+
+// SKIPJobList contains a list of SKIPJob
```
esphen, line 7:
```suggestion
// RoutingList contains a list of Routing
```

README.md
```diff
@@ -262,6 +262,28 @@ spec:
...
+custom resource. Only types are shown here. The fields are documented in the API, see [the API](https://pkg.go.dev/github.com/kartverket/skiperator@v1.0.0/api/v1alpha1)
+```yaml
+apiVersion: skiperator.kartverket.no/v1alpha1
+kind: Routing
+metadata:
+  name: sample-routing
+  namespace: sample
+spec:
+  hostname: app.example.com
+  redirectToHTTPS: true
```
esphen, line 277: Consider not including this line in the example as no sane person would want to expose plain HTTP in 2024

esphen (commented):
Cool! 👏 Some thoughts for discussion below

controllers/routing/certificate.go
```diff
@@ -0,0 +1,106 @@
...
+
+	return requests
+}
+
+func (r *RoutingReconciler) reconcileCertificate(ctx context.Context, routing *skiperatorv1alpha1.Routing) (reconcile.Result, error) {
+	// TODO: Fix controllerProgressing
+
+	var err error
+	//controllerName := "Certificate"
+	//r.SetControllerProgressing(ctx, routing, controllerName)
```
martinhny, line 48: IMO, drop this. IIRC this will change the resourceVersion of the object without actually doing anything which can cause race conditions when we try to apply updates to the object. It's better to only update the status after something went wrong/well.

martinhny (approved):
LGTM. But I feel like this could have been its own operator.

README.md
```diff
@@ -262,6 +262,27 @@ spec:
     restartPolicy: ""
 ```
 
+## Routing reference
```
martinhny, line 265: add a disclaimer not to use 'normal' routing and this at the same time?

eliihen (approved):

PR #424, opened 2024-03-11
Introduce anti-affinity

evenh:
This is an implementation of anti-affinity rules based on hostname and VMware ESXi host labels (if present). It's currently implemented using [Pod Topology Spread Constraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/) but can be switched to using regular (anti-)affinity rules.

### DIFF
--- a/pkg/resourcegenerator/core/pod.go
+++ b/pkg/resourcegenerator/core/pod.go
@@ -20,7 +20,7 @@ func CreatePodSpec(container corev1.Container, volumes []corev1.Volume, serviceA
 		}
 	}
 
-	return corev1.PodSpec{
+	p := corev1.PodSpec{
 		Volumes: volumes,
 		Containers: []corev1.Container{
 			container,
@@ -44,16 +44,16 @@ func CreatePodSpec(container corev1.Container, volumes []corev1.Volume, serviceA
 		ImagePullSecrets:  []corev1.LocalObjectReference{{Name: "github-auth"}},
 		SchedulerName:     corev1.DefaultSchedulerName,
 		PriorityClassName: fmt.Sprintf("skip-%s", priority),
-		Affinity: &corev1.Affinity{
-			PodAntiAffinity: &corev1.PodAntiAffinity{
-				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
-					affinityTermForAppAndKey(container.Name, Hostname),
-					affinityTermForAppAndKey(container.Name, OnPremFailureDomain),
-				},
-			},
-		},
 	}
 
+	if !podSettings.DisablePodSpreadTopologyConstraints {
+		p.TopologySpreadConstraints = []corev1.TopologySpreadConstraint{
+			spreadConstraintForAppAndKey(container.Name, Hostname),
+			spreadConstraintForAppAndKey(container.Name, OnPremFailureDomain),
+		}
+	}
+
+	return p
 }
 
 func CreateApplicationContainer(application *skiperatorv1alpha1.Application, opts PodOpts) corev1.Container {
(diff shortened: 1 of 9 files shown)


pkg/resourcegenerator/core/pod.go
```diff
@@ -44,16 +45,20 @@ func CreatePodSpec(container corev1.Container, volumes []corev1.Volume, serviceA
...
-		},
 	}
 
+	// Global feature flag
+	if !flags.FeatureFlags.DisablePodTopologySpreadConstraints {
+		// Allow override per application
+		if !podSettings.DisablePodSpreadTopologyConstraints {
+			p.TopologySpreadConstraints = []corev1.TopologySpreadConstraint{
+				spreadConstraintForAppAndKey(container.Name, Hostname),
+				spreadConstraintForAppAndKey(container.Name, OnPremFailureDomain),
```
kagkarlsson, line 56: When does the failure-domain stuff kick in? I what scenarios will there be multiple?

pkg/resourcegenerator/core/pod.go
```diff
@@ -212,20 +217,19 @@ func getContainerPorts(application *skiperatorv1alpha1.Application, opts PodOpts
...
+		MaxSkew:           1,
+		TopologyKey:       string(key),
+		WhenUnsatisfiable: corev1.ScheduleAnyway,
+		LabelSelector: &v1.LabelSelector{
+			MatchExpressions: []v1.LabelSelectorRequirement{
+				{
+					Key:      "app",
+					Operator: v1.LabelSelectorOpIn,
+					Values:   []string{appName},
 				},
```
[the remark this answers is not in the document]
evenh: The short answer is no.

Since these topology spread constraints are applied to a `spec`, it is only applied in the namespace specified by the outer resource (`StatefulSet`, `Deployment` etc.)

pkg/resourcegenerator/core/pod.go
evenh: > You can use topology spread constraints to control how Pods are spread across your cluster among failure-domains such as regions, zones, nodes, and other user-defined topology domains. This can help to achieve high availability as well as efficient resource utilization.
> – [K8s doc](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/)

As far as I know, these constraints are are only applied when scheduling pods onto nodes (in addition to CPU/mem requests, priority classes, persistent storage etc.). 

As for multiple failure domains:
- `kubernetes.io/hostname`: We want to spread pods over more nodes in order to spread resource usage and guard against individual node failures (VM down, flaky network etc). This is a K8s standard label.
- `onprem.gke.io/failure-domain-name`: This is an Anthos/GKE On-Prem specific label. In addition to individual VM failures we want to maximize the pod spread over individual physical ESXi hosts (containing one or more VMs acting as K8s nodes).
- [my understanding] If one of the labels is not present on the node it's simply not taken into account, resulting in using only the other specified node labels (in this code, the hostname label). This should be verified by running a build in one of the cloud clusters and see pods be scheduled over the various available hosts. If this understanding is correct, a new PR should be issued to also take the K8s standard label `topology.kubernetes.io/zone` (populated by either the Kubelet or the cloud-controller-manager provided by GKE). It's not implemented today because it's missing in GKE On-Prem.

kagkarlsson: > onprem.gke.io/failure-domain-name: This is an Anthos/GKE On-Prem specific label. In addition to individual VM failures we want to maximize the pod spread over individual physical ESXi hosts (containing one or more VMs acting as K8s nodes).

And the ESXi hosts have different values for this label? Sounds like a good idea!

pkg/resourcegenerator/core/pod.go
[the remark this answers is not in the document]
kagkarlsson: Ah! 💡 👍

pkg/resourcegenerator/core/pod.go
evenh: Yes, they have :-) It's explicitly mentioned in the [documentation](https://cloud.google.com/anthos/clusters/docs/on-prem/latest/concepts/high-availability-disaster-recovery#user_cluster) and can be verified by running this command against an environment, e.g. `prod` or `dev`:
```shell
kubectl get nodes -o json | jq '.items | map({name: .metadata.name, failureDomain: .metadata.labels["onprem.gke.io/failure-domain-name"]}) | group_by(.failureDomain) | map({failureDomain: .[0].failureDomain, nodes: map(.name)})'
```

BardOve (approved):

PR #437, opened 2024-04-02
Fix runtime panic introduced by #436

omaen:
I introduced a runtime panic in #436 because I assumed that the default from the CRD was propogated to the reconciler. Added a test to catch the case.


berg-thom (approved):

controllers/application/service_monitor.go
evenh, line 1: This needs to be applied to pod_monitor.go as well + tests.

evenh (changes requested):

controllers/application/service_monitor.go
evenh: `controllers/skipjob/pod_monitor.go`. It is susceptible to the same nil bug.

evenh (approved):

PR #449, opened 2024-04-22
Kartverket.cloud DNS fix

BardOve:
This PR aims to enable users to force external gateways and network policy settings to statkart.no and kartverket.cloud addresses. This adds Application.Spec.IsInternal, setting this to false will force the use of external istio gateway.


api/v1alpha1/application_types.go
```diff
@@ -68,6 +68,12 @@ type ApplicationSpec struct {
 	//+kubebuilder:validation:Optional
 	Ingresses []string `json:"ingresses,omitempty"`
 
+	// If the application is external, and uses statkart.no or kartverket.cloud name, set this to false. This will 
+	// make the application reachable only from the internet. This is to override the default behaviour of 
+	// setting kartverket.cloud or statkart.no applications as internal.
+	//+kubebuilder:validation:Optional
```
martinhny, line 74:
```suggestion
	//+kubebuilder:validation:Optional
       //+kubebuilder:default:=true
```
and remove `GetIsInternal` function

api/v1alpha1/routing_types.go
```diff
@@ -32,6 +32,9 @@ type RoutingSpec struct {
 	//+kubebuilder:validation:Required
 	Hostname string `json:"hostname"`
 
+	//+kubebuilder:validation:Optional
```
martinhny, line 35:
```suggestion
	//+kubebuilder:validation:Optional
	//+kubebuilder:default:=true
```
and remove GetIsInternal function

BardOve: If this is true by default, it will show as true when using external addresses. The function is to avoid ambiguous config for external addresses.

PR #475, opened 2024-06-10
disable metrics merging when not using custom metrics

evenh:
This change ensures that metrics merging won't be attempted if the `Application` spec doesn't contain explicit Prometheus configuration. In that case only the metrics returned from the istio-proxy sidecar is scraped, and the sidecar won't attempt to scrape the application itself.

Tested in `atkv3-{sandbox,dev}` and verified that the logs containing the following line are absent after the rollout:
> `failed scraping application metrics: error scraping http://localhost:80/metrics:` 

When deploying this change, every application that _don't_ have specified a Prometheus section of their spec will be restarted.


evenh:
I've reverted the `SKIPJob` fix for now as it's not a big problem and the tests break. `Application` and `SKIPJob` might look similar but their implementations differ in a number of ways. Let's address this in the upcoming rewrite.

PR #494, opened 2024-07-22
Fix missing security context for cloudsql proxy

martinhny:
PSA: Restricted stops the pod from working without this


eliihen (approved):
Great 👍

eliihen:
Docs on allowed values are here: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted

PR #507, opened 2024-08-14
Fix custom cert

evenh:


api/v1alpha1/hosts.go
```diff
@@ -52,3 +56,52 @@ func NewHost(hostname string) (*Host, error) {
...
+
+func (hs *HostCollection) AllHosts() []*Host {
+	hosts := make([]*Host, 0, len(hs.hosts))
+	for _, host := range hs.hosts {
+		hosts = append(hosts, host)
+	}
+	return hosts
+}
+
+func (hs *HostCollection) Hostnames() []string {
```
martinhny, line 97: `Go doesn't provide automatic support for getters and setters. There's nothing wrong with providing getters and setters yourself, and it's often appropriate to do so, but it's neither idiomatic nor necessary to put Get into the getter's name. If you have a field called owner (lower case, unexported), the getter method should be called Owner (upper case, exported), not GetOwner.`

TIL

martinhny (approved):

api/v1alpha1/hosts.go
```diff
@@ -52,3 +56,52 @@ func NewHost(hostname string) (*Host, error) {
...
+	case false:
+		fallthrough
+	default:
+		hs.hosts[h.Hostname] = h
+	}
+
+	return nil
+}
+
+func (hs *HostCollection) AllHosts() []*Host {
```
martinhny, line 89: I think `hosts.AllHosts()` is a big confusing. The naming could be better but I can't think of anything. @omaen @BardOve any suggestions?

BardOve (approved):

PR #561, opened 2024-11-18
NAV -> Kartverket

kleivane:
update scope prefix and information about delegated scopes to Kartverket

## Summary by CodeRabbit

- **New Features**
	- Updated documentation for the `applications.skiperator.kartverket.no` resource to clarify access requirements and naming conventions.
	- Enhanced descriptions for integration settings related to Maskinporten, ensuring accurate representation of organizational access.


evenh:
Unfortunately this change won't work, as we're currently embedding NAV's API's from [liberator](https://github.com/nais/liberator). When generating CRD's (`make generate`) these changes would be overwritten.

We should probably fix it upstream.

PR #589, opened 2025-01-21
Add JWT-validation for idporten/maskinporten

larsore:
This PR extends `idporten`-spec and `maskinporten`-spec to include a new block called `requestAuthentication`. It can be enabled like this:
```yaml
apiVersion: skiperator.kartverket.no/v1alpha1
kind: Application
metadata:
  name: application-auth
spec:
  image: image
  port: 8080
  ingresses:
    - example.com
  idporten:
    enabled: true
    integrationType: "api_klient"
    scopes:
      - "openid"
    requestAuthentication:
      enabled: true
```
When enabled, an Istio RequestAuthentication resource is generated to validate all incoming requests with a JWT. The correct issuer-uri, jwks-uri and client-id are retrieved from the kubernetes-`Secret` created by digdirator. If the client registration is done outside the application-manifest, a `secretName` needs to be provided. 
The generated `AuthorizationPolicies` also needs to be modified to ensure that all paths requires the presence of a valid token.

The generated requestauthentication resource looks like this:
```yaml
apiVersion: security.istio.io/v1
kind: RequestAuthentication
metadata:
  name: application-auth-jwt-authn
spec:
  jwtRules:
    - audiences:
        - test-client-id-idporten
      forwardOriginalToken: true
      issuer: https://idporten.no
      jwksUri: https://idporten.no/jwks.json
  selector:
    matchLabels:
      app: application-auth
```

### Additional parameters in `requestAuthentication`
- `secretName`: If an IDPorten or MaskinPorten client is registered outside of the application-manifest, the name of a kubernetes-`Secret` holding issuer-uri, jwks-uri and client-id can be provided.
- `forwardOriginalToken`
- `paths`: Which paths requires a valid JWT. If `paths` is set, all other paths not matching will be open.
- `ignorePaths`: Which paths should not require a valid JWT? If `ignorePaths` is set, all other paths not matching will require a valid JWT.
- `tokenLocation`: Is the JWT located as a BearerToken-cookie or as an Authorization-header?
- `outputClaimToHeaders`: If you wish to copy claims from the JWT and add them as headers to the requests.

## Summary by CodeRabbit

- **New Features**
  - Enhanced JWT authentication options for applications, offering configurable token validation, forwarding, and customizable secure/ignored paths.
  - Support for multiple authentication providers with dynamic security policies and updated naming conventions.
  - Introduction of `RequestAuthentication` and `AuthorizationPolicy` resources for managing access control based on JWT tokens.
  - New methods for managing authentication configurations within applications.
  - New package for handling authentication configurations, including methods for retrieving ignored paths.
  - New `IsRequestAuthEnabled` method in the `ApplicationSpec` struct to evaluate request authentication status.
  - Introduction of new default deny policies and JWT-auth policies for enhanced authorization management.
  - New `Digdirator` package with types and interfaces for managing authentication providers.

- **Improvements**
  - Refined application control processes to better integrate new authentication configurations.
  - Updated configuration schemas and tests to ensure improved security governance and operational consistency.
  - Enhanced deepcopy functionality for authentication-related structures.
  - Added support for retrieving specific client instances from Kubernetes for `IDPorten` and `Maskinporten`.


internal/controllers/common/reconciler.go
```diff
@@ -141,6 +143,179 @@ func (r *ReconcilerBase) IsIstioEnabledForNamespace(ctx context.Context, namespa
 	return exists && len(v) > 0
 }
 
+func (r *ReconcilerBase) GetAuthConfigsForApplication(ctx context.Context, application *v1alpha1.Application) (*[]reconciliation.AuthConfig, error) {
```
martinhny, line 146: if this function is only used for applications, it should be in the `internal/controllers/application.go` file

internal/controllers/common/reconciler.go
```diff
@@ -141,6 +143,179 @@ func (r *ReconcilerBase) IsIstioEnabledForNamespace(ctx context.Context, namespa
...
+	if len(authConfigs) > 0 {
+		return &authConfigs, nil
+	} else {
+		return nil, nil
+	}
+}
+
+func getIdentityProviderInfoWithAuthenticationEnabled(ctx context.Context, application *v1alpha1.Application, k8sClient client.Client) ([]reconciliation.IdentityProviderInfo, error) {
+	var providerInfo []reconciliation.IdentityProviderInfo
+	if application.Spec.IDPorten != nil && application.Spec.IDPorten.Authentication != nil && application.Spec.IDPorten.Authentication.Enabled {
```
martinhny, line 201: maybe move this check to utils, name it IsIDPortenEnabled, for clarity?

internal/controllers/common/reconciler.go
```diff
@@ -141,6 +143,179 @@ func (r *ReconcilerBase) IsIstioEnabledForNamespace(ctx context.Context, namespa
...
+		} else {
+			notPaths = nil
+		}
+		providerInfo = append(providerInfo, reconciliation.IdentityProviderInfo{
+			Provider:   reconciliation.ID_PORTEN,
+			SecretName: *secretName,
+			NotPaths:   notPaths,
+		})
+	}
+	if application.Spec.Maskinporten != nil && application.Spec.Maskinporten.Authentication != nil && application.Spec.Maskinporten.Authentication.Enabled == true {
```
martinhny, line 237: same here, IsMaskinPortenEnabled

larsore: Done👌

internal/controllers/common/reconciler.go
larsore: Done👌

internal/controllers/common/reconciler.go
larsore: Done👌

api/v1alpha1/istiotypes/jwt_authentication.go
```diff
@@ -0,0 +1,57 @@
...
+	// If set to true, the original token will be kept for the upstream request. Default is true.
+	ForwardOriginalToken *bool `json:"forwardOriginalToken,omitempty"`
+
+	// Where to find the JWT in the incoming request
+	//
+	// An enum value of `header` means that the JWT is present in the `Authorization` as a Bearer Token.
+	// An enum value of `cookie` means that the JWT is present as a cookie called `BearerToken`.
+	//
+	// +kubebuilder:validation:Enum=header;cookie
+	TokenLocation *string `json:"tokenLocation,omitempty"`
```
martinhny, line 25: are you sure you don't want a default value, or set this as required?

pkg/reconciliation/reconciliation.go
```diff
@@ -38,6 +39,32 @@ type baseReconciliation struct {
...
+}
+
+var (
+	MASKINPORTEN IdentityProvider = "MASKINPORTEN"
+	ID_PORTEN    IdentityProvider = "ID_PORTEN"
+)
+
+type IdentityProvider string
+
+type IdentityProviderInfo struct {
```
martinhny, line 64: This struct is not used in the reconciliation object, so imo it doesn't belong here. I think you should consider just returning the identity provider and secretname from `getIdentityProviderInfoWithAuthenticationEnabled`

larsore: I agree, but where does it belong?😅

martinhny: top of the file where its used :)

internal/controllers/application.go
```diff
@@ -411,3 +418,176 @@ func validateIngresses(application *skiperatorv1alpha1.Application) error {
...
+	if len(authConfigs) > 0 {
+		return &authConfigs, nil
+	} else {
+		return nil, nil
+	}
+}
+
+func getIdentityProviderInfoWithAuthenticationEnabled(ctx context.Context, application *skiperatorv1alpha1.Application, k8sClient client.Client) ([]IdentityProviderInfo, error) {
+	var providerInfo []IdentityProviderInfo
+	if util.IsIDPortenAuthenticationEnabled(application) {
```
martinhny, line 477: IDPorten/MaskinPorten could implement an interface and I think we could get away with some duplicate code, but thats not a requirement to merge this pr. Only if you want to :P

pkg/resourcegenerator/istio/authorizationpolicy/authorization_policy.go
```diff
@@ -25,68 +26,148 @@ func Generate(r reconciliation.Reconciliation) error {
...
 	}
+	authConfigs := r.GetAuthConfigs()
+	if authConfigs != nil {
+		if len(*authConfigs) > 0 {
+			for _, authConfig := range *authConfigs {
+				if authConfig.NotPaths != nil {
+					allowPaths = append(allowPaths, *authConfig.NotPaths...)
+				}
+			}
+			r.AddResource(
```
martinhny, line 53: The pattern in the other resources have been only one AddResource per file, so following that pattern you should create a new file to create an authorizationpolicy for authConfig

martinhny: Its also to stop the files from getting giant and bloated, it should be easy to understand whats going on

internal/controllers/application.go
larsore: Refactored into using interface for IDPorten/Maskinporten to implement. Placed it in `pkg/jwtAuth`

pkg/jwtAuth/authConfig.go
```diff
@@ -0,0 +1,53 @@
+package jwtAuth
```
martinhny, line 1:
```suggestion
package jwt_auth
```
same for folder name

pkg/jwtAuth/identityProvider.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+type ProviderURIs struct {
+	Provider  IdentityProvider
+	IssuerURI string
+	JwksURI   string
+	ClientID  string
+}
+
+type ProviderOps interface {
```
martinhny, line 24: nice job on the interface, but i think that you should move this to the digdirator directory, and rename it Digdirator (If we implement azure, maybe we rename it to provider)

pkg/jwtAuth/idporten.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (i *IDPortenOps) IsEnabled(application skiperatorv1alpha1.Application) bool {
+	return util.IsIDPortenAuthenticationEnabled(application)
+}
+
+func (i *IDPortenOps) Provider() IdentityProvider {
+	return ID_PORTEN
+}
+
+func (i *IDPortenOps) GetSecret(k8sClient client.Client, ctx context.Context, application skiperatorv1alpha1.Application) (*v1.Secret, error) {
```
martinhny, line 24: move the contents of this file to digdirator/idporten.go

pkg/jwtAuth/maskinporten.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (i *MaskinportenOps) IsEnabled(application skiperatorv1alpha1.Application) bool {
+	return util.IsMaskinPortenAuthenticationEnabled(application)
+}
+
+func (i *MaskinportenOps) Provider() IdentityProvider {
+	return MASKINPORTEN
+}
+
+func (i *MaskinportenOps) GetSecret(k8sClient client.Client, ctx context.Context, application skiperatorv1alpha1.Application) (*v1.Secret, error) {
```
martinhny, line 24: move the contents of this file to digdirator/maskinporten.go

pkg/jwtAuth/authConfig.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"context"
+	"fmt"
+	skiperatorv1alpha1 "github.com/kartverket/skiperator/api/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type AuthConfigs []AuthConfig
```
martinhny, line 10: i think the authconfigs can belong in reconciliation. the functions can be moved back to the application controller

api/v1alpha1/digdirator/idporten.go
```diff
@@ -75,4 +80,69 @@ type IDPorten struct {
...
+
+func (i *IDPortenClient) GetSecretName() string {
+	return i.Client.Spec.SecretName
+}
+
+func (i *IDPortenClient) GetOwnerReferences() []v1.OwnerReference {
+	return i.Client.GetOwnerReferences()
+}
+
+const IDPortenName DigdiratorName = "idporten"
```
martinhny, line 100: variables on top of file

martinhny: and constants are usually in all caps

api/v1alpha1/digdirator/maskinporten.go
```diff
@@ -15,4 +20,69 @@ type Maskinporten struct {
...
+
+func (m *MaskinportenClient) GetOwnerReferences() []v1.OwnerReference {
+	return m.Client.GetOwnerReferences()
+}
+
+func (m *MaskinportenClient) GetSecretName() string {
+	return m.Client.Spec.SecretName
+}
+
+const MaskinPortenName = "maskinporten"
```
martinhny, line 40: top, and in all caps

pkg/reconciliation/reconciliation.go
```diff
@@ -18,6 +20,15 @@ const (
 	RoutingType     ObjectType = "Routing"
 )
 
+type AuthConfigs []AuthConfig
+
+type AuthConfig struct {
```
martinhny, line 25: lets move authconfigs and authconfig to its own package

pkg/reconciliation/reconciliation.go
```diff
@@ -71,3 +84,51 @@ func (b *baseReconciliation) GetRestConfig() *rest.Config {
...
 }
+
+func (b *baseReconciliation) GetAuthConfigs() *AuthConfigs {
+	return b.authConfigs
+}
+
+func (a *AuthConfigs) GetIgnoredPaths() []string {
+	// Her må ignoredPaths utvides KUN hvis en ignorePath ikke inntreffer i andre authConfigs sine paths
+	var ignoredPaths []string
+	if a != nil {
```
martinhny, line 95: i dont have a fix for you right now, but 4 nested for loops is usually a really bad idea, 2 at most is acceptable

pkg/resourcegenerator/istio/authorizationpolicy/default_deny/authorization_policy.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+	skiperatorv1alpha1 "github.com/kartverket/skiperator/api/v1alpha1"
+	"github.com/kartverket/skiperator/pkg/reconciliation"
+	"github.com/kartverket/skiperator/pkg/resourcegenerator/istio/authorizationpolicy"
+	securityv1api "istio.io/api/security/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func Generate(r reconciliation.Reconciliation) error {
```
martinhny, line 12: is this always generated, if nothing is specified?
what are the conditions to generate this?

larsore: It is not generated if `application.Spec.AuthorizationSettings.AllowAll == true`

pkg/auth/auth_config.go
```diff
@@ -0,0 +1,58 @@
...
+type AuthConfigs []AuthConfig
+
+type AuthConfig struct {
+	Spec         istiotypes.Authentication
+	Paths        []string
+	IgnorePaths  []string
+	ProviderURIs digdirator.DigdiratorURIs
+}
+
+func (authConfigs *AuthConfigs) GetIgnoredPaths() []string {
```
martinhny, line 18:
```suggestion
func (authConfigs *AuthConfigs) GetIgnoredPaths() []string {
	var ignoredPaths map[string]string
	var allowPaths map[string]string
	if authConfigs != nil {
		for _, config := range *authConfigs {
			for _, ignoredPath := range config.IgnorePaths {
				ignoredPaths[ignoredPath] = ignoredPath
			}
			for _, allowPath := range config.Paths {
				allowPaths[allowPath] = allowPath
			}
		}

		for _, path := range allowPaths {
			if _, ok := ignoredPaths[path]; ok {
				delete(ignoredPaths, path)
			}

		}
	}
	return maps.Values(ignoredPaths)
}
```

refactored suggestion using maps instead

pkg/auth/auth_config.go
```diff
@@ -0,0 +1,57 @@
...
+			if _, ok := ignoredPaths[path]; ok {
+				delete(ignoredPaths, path)
+			}
+
+		}
+	}
+	return maps.Values(ignoredPaths)
+}
+
+func (authConfigs *AuthConfigs) IgnorePathsFromOtherAuthConfigs() {
```
martinhny, line 42: Why do we do this? 
Is it to help user from misconfiguring? I'm not sure if we want to do that..

larsore: It is for the case where authentication is configured for two providers. When the auth config for one provider requires authentication on a path, the auth config for the other provider should not override this requirement by setting the path as ignored. For example if maskinporten auth requires `/secureMaskinPorten` to be authenticated, the idea is to enforce authentication even if idporten auth ignores authentication for `/secureMaskinPorten`.

config/crd/skiperator.kartverket.no_applications.yaml
```diff
@@ -553,6 +553,83 @@ spec:
                     maximum: 3600
                     minimum: 1
                     type: integer
+                  authentication:
```
okpedersen, line 556: I feel `authentication` is ambiguous 🤔 E.g., it could be read as some sort of authentication towards IDPorten, or digdir?

Names along the lines of `requestAuthentication`, `builtinAuthentication`, `tokenValidation` or similar could be more readable for users reading the spec.

E.g. `idporten.token_validation.enabled` makes more sense to me than `idporten.authentication.enabled`.

config/crd/skiperator.kartverket.no_applications.yaml
```diff
@@ -553,6 +553,83 @@ spec:
...
+                    description: Authentication specifies how incoming JWT's should
```
okpedersen, line 557: Nitpick: Use plural and not possessive form:
```suggestion
                    description: Authentication specifies how incoming JWTs should
```

config/crd/skiperator.kartverket.no_applications.yaml
```diff
@@ -553,6 +553,83 @@ spec:
...
+                      be validated.
+                    properties:
+                      enabled:
+                        description: |-
+                          Whether to enable JWT validation.
+                          If enabled, incoming JWT's will be validated against the issuer specified in the app registration and the generated audience.
```
okpedersen, line 563: Nitpick: possessive vs plural
```suggestion
                         If enabled, incoming JWTs will be validated against the issuer specified in the app registration and the generated audience.
Comment
```

config/crd/skiperator.kartverket.no_applications.yaml
```diff
@@ -553,6 +553,83 @@ spec:
...
+                        type: boolean
+                      forwardOriginalToken:
+                        default: true
+                        description: If set to `true`, the original token will be
+                          kept for the upstream request. Defaults to `true`.
+                        type: boolean
```
okpedersen, line 569: This is the only place using "token" and not JWT. Language should be standardised (i.e., prefer JWT). E.g., rename to `forwardOriginalJwt` and description to match.

Also, the use of "original" can imply that an "unoriginal" (derived, exchanged, etc) token exists. If that is not the case, maybe we can remove "original" (e.g., `forwardToken` or `forwardJwt`)?

config/crd/skiperator.kartverket.no_applications.yaml
```diff
@@ -553,6 +553,83 @@ spec:
...
+                      ignorePaths:
+                        description: |-
+                          IgnorePaths specifies paths that do not require an authenticated JWT.
+
+                          The specified paths must start with '/'.
```
okpedersen, line 574: Would be nice with an example in this description 🤔  E.g. `/auth/login`, just to show users how it works

okpedersen: Also, is it exact matches or prefix matches? How does it handle ending slashes?

Example, given `/auth/login` is specified, how does it handle the following examples:
`/auth/login` (exact match, so should be ignored)
`/auth/login/` (ending slash)
`/auth/login/idporten` (prefix match)

config/crd/skiperator.kartverket.no_applications.yaml
```diff
@@ -553,6 +553,83 @@ spec:
...
+                        items:
+                          type: string
+                        maxItems: 50
+                        type: array
+                        x-kubernetes-list-type: set
+                      outputClaimToHeaders:
+                        description: "This field specifies a list of operations to
+                          copy the claim to HTTP headers on a successfully verified
+                          token.\nThe header specified in each operation in the list
+                          must be unique. Nested claims of type string/int/bool is
```
okpedersen, line 584: What does "nested claims for type string/int/bool" mean, in practice?

config/crd/skiperator.kartverket.no_applications.yaml
```diff
@@ -553,6 +553,83 @@ spec:
...
+                          header: x-test-environment-flag\n\t  claim: test-flag\n\t-
+                          header: x-jwt-claim-group\n\t  claim: nested.key.group\n\n```"
+                        items:
+                          properties:
+                            claim:
+                              description: The claim to be copied.
+                              type: string
+                            header:
+                              description: The name of the HTTP-header for which the
+                                specified claim will be copied to.
```
okpedersen, line 596: Nitpick:
```suggestion
                              description: The name of the HTTP header the specified claim will be copied to.
```

config/crd/skiperator.kartverket.no_applications.yaml
```diff
@@ -553,6 +553,83 @@ spec:
...
+                              type: string
+                          required:
+                          - claim
+                          - header
+                          type: object
+                        type: array
+                      paths:
```
okpedersen, line 603: How does `paths` work vs `ignoredPaths`? Ordering, precedence, etc 🤔

config/crd/skiperator.kartverket.no_applications.yaml
```diff
@@ -553,6 +553,83 @@ spec:
...
+
+                          The specified paths must start with '/'.
+                        items:
+                          type: string
+                        maxItems: 50
+                        type: array
+                        x-kubernetes-list-type: set
+                      secretName:
+                        description: |-
+                          The name of the kubernetes Secret containing OAuth2-credentials.
```
okpedersen, line 615: Nitpick:
```suggestion              
                          The name of the Kubernetes Secret containing OAuth2 credentials.
```

config/crd/skiperator.kartverket.no_applications.yaml
```diff
@@ -553,6 +553,83 @@ spec:
...
+
+                          If omitted, the associated client registration in the application manifest is used for JWT validation.
+                        type: string
+                      tokenLocation:
+                        default: header
+                        description: |-
+                          Where to find the JWT in the incoming request
+
+                          An enum value of `header` means that the JWT is present in the `Authorization`-header as a Bearer Token.
```
okpedersen, line 624: Nitpick:
```suggestion
                          An enum value of `header` means that the JWT is present in the `Authorization` header as a `Bearer` token.
```

config/crd/skiperator.kartverket.no_applications.yaml
```diff
@@ -756,6 +833,83 @@ spec:
               maskinporten:
                 description: Settings for Maskinporten integration with Digitaliseringsdirektoratet
                 properties:
+                  authentication:
```
okpedersen, line 836: See comments for `idporten`, most of them apply here too.

pkg/auth/auth_config.go
martinhny: @omaen do we want this behaviour?

config/crd/skiperator.kartverket.no_applications.yaml
larsore: I agree, I think `requestAuthentication` or `tokenValidation` is more fitting.

config/crd/skiperator.kartverket.no_applications.yaml
larsore: I agree. Changing it to `forwardJwt`👍

config/crd/skiperator.kartverket.no_applications.yaml
larsore: It is exact match but supports wildcard by using `*`. I can update the description to be more informative about this.

martinhny (approved):

larsore:
@coderabbitai full review

larsore:
@coderabbitai summary

PR #593, opened 2025-02-03
Add support for profiling via pprof

evenh:
Disabled by default in case we're incurring a performance hit

## Summary by CodeRabbit

- **New Features**
  - Introduced performance profiling support for enhanced monitoring and diagnostics.
  - Profiling can now be activated via an environment configuration, enabling a dedicated endpoint for runtime performance insight.

### DIFF
--- a/cmd/skiperator/main.go
+++ b/cmd/skiperator/main.go
@@ -3,6 +3,9 @@ package main
 import (
 	"flag"
 	"fmt"
+	"os"
+	"strings"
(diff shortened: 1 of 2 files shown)


evenh:
That's correct @omaen. It is only accessible if we port-forward to that port or expose it via a `Service`/Istio – which we don't do today.

martinhny (approved):
cool

PR #596, opened 2025-02-06
SKIP-1512 - Generate network policy if external rule is ip

martinhny:
## Summary by CodeRabbit

- **Style**
  - Updated an error message for a more consistent presentation when finalizing applications.

- **New Features**
  - Enhanced processing for external network access by adding IP validation and port mapping for outbound rules.
  - Introduced a new application configuration that specifies external access with defined host and port rules.

- **Tests**
  - Expanded test coverage by adding scenarios to validate the behavior of external network access policies, including new assertions for the `external-ip-policy`.
  - Added a new test step to evaluate the application of the external IP policy.

### DIFF
--- a/pkg/resourcegenerator/networkpolicy/dynamic/common.go
+++ b/pkg/resourcegenerator/networkpolicy/dynamic/common.go
@@ -5,9 +5,11 @@ import (
 	"github.com/kartverket/skiperator/api/v1alpha1/podtypes"
 	"github.com/kartverket/skiperator/pkg/reconciliation"
 	"github.com/kartverket/skiperator/pkg/util"
+	v1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"regexp"
 	"slices"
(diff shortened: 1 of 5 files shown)


pkg/resourcegenerator/networkpolicy/dynamic/common.go
```diff
@@ -93,9 +97,41 @@ func getEgressRules(accessPolicy *podtypes.AccessPolicy, appNamespace string) []
 		egressRules = append(egressRules, getEgressRule(rule, appNamespace))
 	}
 
+	for _, externalRule := range accessPolicy.Outbound.External {
+		if externalRule.Ports == nil || externalRule.Ip == "" || !ipAddress.MatchString(externalRule.Ip) {
```
evenh, line 101:
```suggestion
		if externalRule.Ports == nil || externalRule.Ip == "" || net.ParseIP(externalRule.Ip) == nil {
```

evenh:
Looks good to me. Added an alternative approach that's more idiomatic if you'd like.

pkg/resourcegenerator/networkpolicy/dynamic/common.go
martinhny: even better, thanks :smile:

Griznah (approved):

PR #615, opened 2025-02-27
Bugfix JWT-auth

larsore:
Applying the following skiperator application should allow unauthenticated requests to `/open` while require auth on all other paths.
```yaml
apiVersion: skiperator.kartverket.no/v1alpha1
kind: Application
metadata:
  name: application
spec:
  image: image
  port: 8080  
  idporten:
    enabled: true
    integrationType: "api_klient"
    scopes:
      - "openid"
    requestAuthentication:
      enabled: true
      ignorePaths: 
        - /open
```
The result was that an authorizationpolicy that allowed requests to `/open` was not created since allowed paths from idporten/maskinporten were only included if `spec.authorizationSettings != nil`.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the authorization behavior by ensuring that all designated ignored paths are always consistently applied, regardless of other configuration settings. This update enhances the reliability of how authorization policies process request paths.
	- Adjusted access control rules by removing the `/publicEndpoint` from the allowed operations in `application-one`, while retaining access to `/loginIdPorten`.
	- Removed the `authorizationSettings` section related to public endpoint access in `application-one`.
	- Restored `/publicEndpoint` back into the allowed paths for `application-two`.

### DIFF
--- a/pkg/resourcegenerator/istio/authorizationpolicy/allow/authorization_policy.go
+++ b/pkg/resourcegenerator/istio/authorizationpolicy/allow/authorization_policy.go
@@ -27,10 +27,10 @@ func Generate(r reconciliation.Reconciliation) error {
 	}
 
 	var allowedPaths []string
+	// Include ignored paths from auth config as they should be accessible without authentication
+	allowedPaths = append(allowedPaths, r.GetAuthConfigs().GetIgnoredPaths()...)
 	if application.Spec.AuthorizationSettings != nil {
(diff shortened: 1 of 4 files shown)


martinhny (approved):
should add a test

larsore:
@martinhny I updated one test under `jwt-auth`. Removed `authorizationSettings.allowList` as this is already tested in another test.

PR #639, opened 2025-03-27
Bump to v3

martinhny:
To release packages we need to upgrade the modules field in go.mod, however https://go.dev/wiki/Modules#releasing-modules-v2-or-higher suggests that when you do this for the first time you should create a new version as well, so we should bump skiperator to v3.0.0 after this PR.
Also bumped dependencies and go version.


.github/workflows/build-and-deploy.yaml
evenh, line 1: Maybe we should use https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file for all workflows?

PR #649, opened 2025-04-07
Fix infinite reconcile loop

evenh:
Could not reproduce locally, so I added a condition to catch this misconfiguration.

SKIP-1201

## Summary by CodeRabbit

- **New Features**
  - Enhanced access policy validation now includes external rule checks with improved status reporting.
  - Introduced an optional IP field for external rules to support TCP-based access configurations.

- **Documentation**
  - Updated resource descriptions to clarify host property usage by specifying that hostnames exclude subdomains.

- **Tests**
  - Expanded test coverage to verify the new external rule conditions and ensure accurate configuration status reporting.

### DIFF
--- a/internal/controllers/application.go
+++ b/internal/controllers/application.go
@@ -3,6 +3,9 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"regexp"
+	"strings"
+
 	certmanagerv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	skiperatorv1alpha1 "github.com/kartverket/skiperator/api/v1alpha1"
 	"github.com/kartverket/skiperator/api/v1alpha1/digdirator"
@@ -47,7 +50,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation/field"
-	"regexp"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -55,7 +57,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	"strings"
 )
 
 // +kubebuilder:rbac:groups=skiperator.kartverket.no,resources=applications;applications/status,verbs=get;list;watch;update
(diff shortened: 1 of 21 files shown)


pkg/resourcegenerator/maskinporten/maskinporten.go
```diff
@@ -23,7 +24,7 @@ func Generate(r reconciliation.Reconciliation) error {
 	}
 
 	if !MaskinportenSpecifiedInSpec(application.Spec.Maskinporten) {
-		ctxLog.Info("Maskinporten not specified in spec, skipping generation")
+		ctxLog.Debug("Maskinporten not specified in spec, skipping generation")
```
evenh, line 27: Unrelated change, but changed to debug to follow every other generator so it can be filtered out if necessary

internal/controllers/common/util.go
```diff
@@ -44,6 +48,32 @@ func IsInternalRulesValid(accessPolicy *podtypes.AccessPolicy) bool {
...
+		if len(rule.Host) == 0 {
+			return false
+		}
+
+		normalizedHost := strings.ToLower(rule.Host)
+		if slices.Contains(seenHosts, normalizedHost) {
+			return false
+		}
+
+		if err := domain.Check(normalizedHost); err != nil {
```
okpedersen, line 67: Do we have a test covering this condition?

okpedersen (approved):
Looks good to me. One question, though 🤔

okpedersen (approved):

internal/controllers/common/util.go
martinhny, line 67: When adding cloudsql hosts or other external hosts where we only care about the IP then this will break many existing rules, since the host name can be set as anything

internal/controllers/application.go
```diff
@@ -262,13 +263,19 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req reconcile.Req
...
+	} else if accessPolicy != nil && !common.IsExternalRulesValid(accessPolicy) {
+		conditions = append(conditions, common.GetExternalRulesCondition(app, metav1.ConditionFalse))
+		app.Status.AccessPolicies = skiperatorv1alpha1.INVALIDCONFIG
 	} else {
 		conditions = append(conditions, common.GetInternalRulesCondition(app, metav1.ConditionTrue))
+		conditions = append(conditions, common.GetExternalRulesCondition(app, metav1.ConditionTrue))
 		app.Status.AccessPolicies = skiperatorv1alpha1.READY
 	}
+
 	app.Status.Conditions = conditions
```
martinhny, line 279: feel like we could rework the logic here a bit

tests/application/access-policy/only-unique-hosts-assert.yaml
```diff
@@ -0,0 +1,55 @@
...
+  accessPolicy:
+    outbound:
+      external:
+        - host: somehost.com
+          ip: 100.100.100.42
+          ports:
+            - name: sftp
+              port: 22
+              protocol: TCP
+        - host: somehost.com
```
martinhny, line 17: not every host has a domain name

evenh: Handled in 07e2d35

internal/controllers/common/util.go
evenh: Covered with tests

PR #696, opened 2025-06-27
Add feature flag to enable locally built images

larsore:
We would like the opportunity to run our locally built docker images in a Skiperator-application on a local kind-cluster. This PR adds a feature flag for this which disables resolving image tag for the deployment and sets the `imagePullPolicy`in the container to `Never`.

## Summary by CodeRabbit

* **New Features**
  * Added a feature flag to control support for locally built images.
  * When enabled, image pull policy is set to never pull images, and image tag resolution is skipped.

* **Behavior Changes**
  * Image pull and resolution logic now adapts based on the feature flag, allowing improved support for local development workflows.

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -15,12 +15,14 @@ var FeatureFlags *Features
 type Features struct {
 	DisablePodTopologySpreadConstraints bool
 	EnableProfiling                     bool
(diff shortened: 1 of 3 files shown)


evenh (approved):
LGTM. Any objections @BardOve / @omaen?

PR #702, opened 2025-07-11
SKIP-1704: exception to egress rules

berg-thom:
Would like inputs on how to test this if possible - I added an exception to the IP block assert in default-deny-network test, but it seems to run fine even without this exception, so it seems Chainsaw does not look at all the fields when asserting. I tried digging into Chainsaw for config options or alternate ways to set up a test, but could not get it working.
Also, a negative test with missing/invalid configmap would result in Skiperator not starting at all, so it could not be tested with a regular setup. We could, I suppose, create a separate entry in the Makefile with an invalid config and assert an error message in some way, but am unsure whether this is the right approach.
As it is, this change has been tested manually and found to produce the expected change to default-deny configmap.

## Summary by CodeRabbit

* **New Features**
  * Introduced support for loading and validating cluster network CIDR configurations from a Kubernetes ConfigMap before controller startup.
  * Default deny network policies are now dynamically generated based on cluster CIDR settings, enhancing network policy enforcement.

* **Bug Fixes**
  * Improved validation to ensure only valid and non-empty CIDR blocks are accepted in cluster configurations.

* **Tests**
  * Added unit tests to verify correct parsing and validation of cluster CIDR configurations from ConfigMaps.
  * Updated test and sample ConfigMaps to reflect new cluster CIDR configuration requirements.


cmd/skiperator/main.go
```diff
@@ -119,6 +128,24 @@ func main() {
...
+		setupLog.Error(err, "unable to get configmap skip-cluster-node-cidr")
+		os.Exit(1)
+	}
+
+	skipClusterList, err := createSKIPClusterListFromConfigMap(&clusterConfigMap)
+	if err != nil {
+		setupLog.Error(err, "unable to create SKIPClusterList", "controller", "SKIPClusterList")
+		os.Exit(1)
+	}
+
```
martinhny, line 148:
```suggestion
	var skipClusterList *util.SKIPClusterList
	configCheckClient, err := client.New(kubeconfig, client.Options{})
	err = skipClusterList.LoadConfigFromConfigMap(configCheckClient)
	if err != nil {
		setupLog.Error(err, "couldnt log config blabla")
		os.Exit(1)
	}


```

pkg/util/skipcluster.go
```diff
@@ -0,0 +1,24 @@
+package util
```
martinhny, line 1: maybe put this file in `internal/config` or something? i think it doesn't belong in pkg, since its not used by the controllers and its not really for external use

tests/cluster-config/skip-cluster-node-cidr.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
martinhny, line 1: for integration testing, i think this is enough. 

but i think it would be useful to add some unit tests to verify the logic

internal/config/skipcluster.go
```diff
@@ -0,0 +1,105 @@
...
+	Name              string   `yaml:"name"`
+	ControlPlaneCIDRs []string `yaml:"controlPlaneCIDRs"`
+	WorkerNodeCIDRs   []string `yaml:"workerNodeCIDRs"`
+}
+
+type SKIPClusterList struct {
+	Clusters []*SKIPCluster `yaml:"clusters"`
+}
+
+func ClusterList(clusters ...*SKIPCluster) *SKIPClusterList {
```
martinhny, line 24: unused func

internal/config/skipcluster.go
```diff
@@ -0,0 +1,107 @@
...
+	// check for empty strings and validate that the strings are valid CIDRs
+	for _, element := range skipClusterList.Clusters {
+		err := checkCIDRsAreNotEmpty(element)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	for _, element := range skipClusterList.CombinedCIDRS() {
+		err := checkValidCIDR(element)
```
BardOve, line 76: same
```suggestion
		err = checkValidCIDR(element)
```

BardOve (approved):

PR #703, opened 2025-07-17
skip-1725: backoff limit

evenboee:
previously did not differentiate between nil and 0 for backofflimit. without extensive knowledge of mergo it was easier to just set defaults the old school way. now explicit 0 for backofflimit is supported

## Summary by CodeRabbit

* **Refactor**
  * Improved reliability of default value assignment for jobs by updating how defaults are set.
  * Removed an external dependency to streamline the application.

* **Chores**
  * Cleaned up unused dependencies for a leaner codebase.

### DIFF
--- a/api/v1alpha1/skipjob_types.go
+++ b/api/v1alpha1/skipjob_types.go
@@ -5,7 +5,6 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


SnorreSelmer (approved):
Ser bra ut

PR #733, opened 2025-08-26
update generate to use go tool + docs

evenboee:
add clarification to IntOrString with example. 

had some problems with installing controller-gen so i added it and chainsaw as go1.24 tool

## Summary by CodeRabbit

* **Documentation**
  * Updated retry policy docs: default retried error codes no longer list 5xx; note that mixed types (e.g., [503, "retriable-4xx"]) are supported.

* **Chores**
  * Switched generation/test tooling to use the Go toolchain (go tool) and removed the separate tools bootstrap.
  * Added tool declarations to module file and removed tools.go.
  * Simplified Makefile targets and removed GOBIN/PATH exports; bumped/reclassified several module dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,6 @@ SHELL = bash
 .DEFAULT_GOAL = build
 
 $(shell mkdir -p bin)
-export GOBIN = $(realpath bin)
-export PATH := $(GOBIN):$(PATH)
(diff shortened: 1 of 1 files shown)


evenh (approved):
Nice to finally adopt `go tool` 👏🏻

PR #834, opened 2026-03-04
Skip 1031 subPath filesfrom

Morisman96:
### Summary
Adds subPath functionality to filesFrom, allowing users to mount individual files from within a volume into a specific folder in the container without deleting or overwriting existing files in the target mount path. 

#### Changes
- Added subPath field to filesFrom struct
- Refactored  GetContainerVolumeandPodVolumes to two functions, one responsible for setting volume mount and the other for setting volumes. Now Multiple entries can point to the same volume if
- Added and fixed CEL validation for filesFrom, now manifests without exactly on volume type in a filesFrom entry will be rejected by kubeAPI and volume types field must be at least one character long to avoid eternal diff reconcile bug if a volume type is put as "" in the manifest.
-  Added prefixes to volume and volumemount `name `in deployment based on volume type on creation, this is to link the the right volume with the right volumemount and  avoiding naming conflict. `tmp` i excluded from the prefix

## Summary by CodeRabbit

* **New Features**
  * Added `subPath` field to FilesFrom for mounting specific paths within volumes.
  * Enhanced container image validation at reconciliation time.

* **Bug Fixes**
  * Corrected DefaultMode validation range from 0-777 to 0-511 (octal).
  * Improved error messages with detailed categorization for better troubleshooting.

* **Validation Improvements**
  * FilesFrom now requires exactly one volume source (configMap, secret, emptyDir, or persistentVolumeClaim).
  * Added volume name collision detection across different source types.


internal/controllers/common/util_test.go
```diff
@@ -219,5 +219,78 @@ func TestValidateContainerImageString(t *testing.T) {
...
+			{
+				MountPath: "/secret",
+				Secret:    "shared-name",
+			},
+			{
+				MountPath:             "/pvc",
+				PersistentVolumeClaim: "shared-name",
+			},
+		}))
+	},
```
evenh: Vi bør tillatte at ressurser av ulike typer (Secret og PVC) har samme navn

internal/controllers/application.go
```diff
@@ -210,6 +210,11 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req reconcile.Req
 		return common.DoNotRequeue()
 	}
 
+	if err := common.ValidateFilesFrom(application.Spec.FilesFrom); err != nil {
+		rLog.Error(err, "invalid filesFrom in application manifest")
+		r.SetErrorState(ctx, application, err, "invalid filesFrom in application manifest", "InvalidApplication")
```
evenh:
```suggestion
		r.SetErrorState(ctx, application, err, "invalid filesFrom in application manifest", "InvalidApplication")
		return common.DoNotRequeue()
```

evenh (changes requested):

evenh (dismissed):

BardOve (dismissed):

evenh (approved):

BardOve (approved):

PR #966, opened 2026-09-04
Keep custom TLS certificates in istio-gateways under standard routing

evenh:
Legacy routing terminated TLS on an Istio Gateway and read the credential from istio-gateways. Standard routing terminates on a ListenerSet in the object's own namespace, and a certificateRef without a namespace resolves there. A team with a custom certificate therefore had to copy the Secret into its own namespace before it could set routingProvider: Standard. That is work legacy routing never asked for, and a migration that silently serves nothing if the team misses it.

Point the listener at istio-gateways instead. Gateway API allows a cross-namespace certificateRef when a ReferenceGrant in the target namespace permits it, so the generator emits one grant per ListenerSet. The grant names the Secret, so it does not open the other certificates in that namespace. A managed certificate is issued next to the ListenerSet and needs no grant. Verified on gateway-api 1.5.1 with Istio 1.30.4: the listener reports ResolvedRefs=True, and Envoy presents the certificate from istio-gateways for that hostname.

Shared Routing is unaffected. Its ListenerSet already lives in istio-gateways, so the reference stays in one namespace and needs no grant. A shared Routing cannot carry a custom certificate at all, because the CRD rejects it.

Readiness probes a custom certificate in istio-gateways and reports a missing or unusable one as CustomCertificateMissing. No wait resolves that condition, because the team must provision the Secret, so it is separated from the dependencies Skiperator is only waiting for. Legacy routing keeps serving the hostname while it holds.

The ReferenceGrant needs no new lifecycle code. The resource processor already lists every registered schema in istio-gateways by object label, which is the same path the legacy Certificate uses.

## Changelog

- Keep custom TLS certificates in `istio-gateways` during standard routing migration.
- Generate a uniquely named, Secret-scoped `ReferenceGrant` for cross-namespace certificate access.
- Report missing or unusable custom certificates as `CustomCertificateMissing`.
- Keep legacy routing active while custom certificates are unavailable.
- Preserve specific API error reasons for readiness failures.
- Add Gateway API RBAC permissions and resource tracking for `ReferenceGrant`.
- Update Gateway API documentation and migration fixtures.
- Add tests for certificate blockers, ReferenceGrants, readiness errors, migration behavior, and cleanup.


andtag (dismissed):

andtag (dismissed):

pkg/gwapi/migration.go
```diff
@@ -15,6 +15,9 @@ const (
 	migrationStalledReason        = commontypes.MigrationStalledReason
 	legacyRoutingActiveReason     = "LegacyRoutingActive"
 	legacyRoutingInactiveReason   = "LegacyRoutingInactive"
+	// customCertificateMissingReason marks a blocker that no wait resolves.
+	// The team must put the custom certificate in the application namespace.
```
BardOve:
```suggestion
```

BardOve (approved):

Morisman96 (approved):

PR #975, opened 2026-09-15
feat(skipjob): support extraContainers on v1beta1

evenh:
Add spec.extraContainers to SKIPJob v1beta1, reusing the ContainerSpec type and pod.CreateExtraContainers already shared with Application.

Entries must set type: init, which produces a native sidecar (an init container with restartPolicy: Always). Kubernetes stops a native sidecar when the job container exits, so the Job can complete. A standard sidecar never exits on its own and would keep the Job running until its deadline, so CEL rejects it. ingressPort is rejected too, since a SKIPJob has no Service.

The field is deliberately v1beta1-only, to give teams a reason to migrate off the deprecated v1alpha1 schema. A CEL rule on the v1alpha1 schema cannot enforce that: the mutating webhook is registered on v1beta1 with matchPolicy Equivalent, so every v1alpha1 write round-trips through the conversion webhook before CEL validation runs, and ConvertFrom drops the field on the way back. The rejection therefore lives in ConvertTo, and v1alpha1 declares the field only so structural pruning does not remove it before conversion sees it.

validateExtraContainers moves to internal/controllers/common so the SKIPJob and Application controllers share it.

SKIP-1895

## Changelog

- Added `spec.extraContainers` support for `SKIPJob` `v1beta1`.
- Added native sidecars through `type: init`.
- Rejected standard sidecars and `ingressPort` for `SKIPJob`.
- Declared `extraContainers` in `v1alpha1` to prevent pruning, while returning a migration error when used.
- Reused `ContainerSpec` and `pod.CreateExtraContainers`.
- Shared extra-container validation between controllers.
- Added CRD documentation and Chainsaw test fixtures.


evenh:
@coderabbitai 

extraContainers is v1beta1-only on purpose. SKIPJob v1alpha1 is deprecated (it already carries deprecated: true and a deprecationWarning), and gating new features behind v1beta1 is how we give teams a concrete reason to migrate. Making the conversion lossless would make the feature fully usable from v1alpha1 and remove that reason entirely.

api/v1beta1/skipjob_types.go
```diff
@@ -64,6 +64,10 @@ type SKIPJobList struct {
...
+// +kubebuilder:validation:XValidation:rule="has(self.cron) || ((has(oldSelf.extraContainers) == has(self.extraContainers)) && (!has(self.extraContainers) || oldSelf.extraContainers == self.extraContainers))",message="extraContainers is immutable for one-off jobs. Please delete your SKIPJob to change it."
+// +kubebuilder:validation:XValidation:rule="!has(self.extraContainers) || self.extraContainers.all(c, self.extraContainers.filter(x, x.name == c.name).size() == 1)",message="extraContainers names must be unique"
+// +kubebuilder:validation:XValidation:rule="!has(self.extraContainers) || self.extraContainers.all(c, !has(c.ingressPort))",message="ingressPort is not supported for SKIPJob"
+// +kubebuilder:validation:XValidation:rule="!has(self.extraContainers) || self.extraContainers.all(c, has(c.type) && c.type == 'init')",message="extraContainers in a SKIPJob must set type: init; a standard sidecar never exits and the Job would never complete"
 type SKIPJobSpec struct {
```
berg-thom, line 71: Lite hjertesukk: Kubebuilder-linjene her begynner å bli latterlig omfattende og borderline uleselig

berg-thom (dismissed):

BardOve (approved):

berg-thom (approved):

